PR #15, opened 2022-07-18
Issue14 rcpp enhancement

jianxiaoyang:


R/simPWSurv.R
```diff
@@ -117,14 +117,14 @@ simPWSurv <- function(n=100,
...
-    utr <- unique(x$Treatment)
     usr <- unique(x$Stratum)
+    utr <- unique(x$Treatment)
     x$failTime <- 0
     x$dropoutTime <- 0
     for(sr in usr){for(tr in utr){
       indx <- x$Stratum==sr & x$Treatment==tr
-      x$failTime[indx] <- rpwexp(n=sum(indx),failRates=filter(failRates,Stratum==sr&Treatment==tr))
-      x$dropoutTime[indx] <- rpwexp(n=sum(indx),failRates=filter(dropoutRates,Stratum==sr&Treatment==tr))
+      x$failTime[indx] <- rpwexpinvRcpp(n=sum(indx),failRates=failRates[failRates$Stratum==sr&failRates$Treatment==tr,])
```
nanxstats, line 126: You probably want to use `drop = FALSE` in all places where the `[` subsetting operator is used: https://adv-r.hadley.nz/subsetting.html#simplify-preserve

R/simfix.R
```diff
@@ -40,6 +41,7 @@ NULL
 #' @param rg As in `simtrial::tenFHCorr()`.
 #' A \code{tibble} with variables \code{rho} and \code{gamma}, both greater than equal
 #' to zero, to specify one Fleming-Harrington weighted logrank test per row.
+#' @param setSeed Bool: Set seed for simulation?
```
nanxstats, line 44: One usually wants to set seed for reproducibility. I guess it would be a more meaningful argument if this is an integer input so that people can **control** what (initial) seed is used.

nanxstats (changes requested):

nanxstats (approved):

nanxstats:
I think this looks good to me now. Any other thoughts? @LittleBeannie

jianxiaoyang:
I just realized maybe we should add a vignette for the parallel computing tutorial (not the benchmark) with a small example (~100 simulations)?

LittleBeannie:
> I just realized maybe we should add a vignette for the parallel computing tutorial (not the benchmark) with a small example (~100 simulations)?

If the tutorial example is complicated, please add a vignette. Otherwise, an example in the spec of `simfix` is sufficient.

LittleBeannie:
> I think this looks good to me now. Any other thoughts? @LittleBeannie

Thanks for the review, @nanxstats . I would like to take a close look at the maxcombo tests (mentioned by Jianxiao) tomorrow and will merge it later.

Issue #25, opened 2023-01-11
API update


LittleBeannie:
In `simtrial`, most of the function/variable/argument name follows the camel case (e.g., the `enrollRates` argument in the `simfix` function). In modern package development, people follow snake case. For example,  `enrollRates` should be `enroll_rate` in snake case. The differences between snake case and camal are summarized in https://www.freecodecamp.org/news/snake-case-vs-camel-case-vs-pascal-case-vs-kebab-case-whats-the-difference/. 

**Function name update:** 
- [x] `fixedBlockRand` -> `randomize_by_fixed_block` @lili-ling-msd, see #32 
- [x] `cutData` -> `cut_data_by_date` @XintongLi2023, see #33
- [x] `cutDataAtCount` -> `cut_data_by_event` @XintongLi2023, see #34
- [x] `getCutDateForCount` -> `get_cut_date_by_event` @XintongLi2023, see #35
- [x] `rpwenroll` -> `rpw_enroll` @lili-ling-msd, see #36
- [x]  `pwexpfit` -> `fit_pwexp` @lili-ling-msd, see #37
- [x] `pMaxCombo` -> `pvalue_maxcombo` @lili-ling-msd, see #38
- [x] `tenFH` -> `wlr` @lili-ling-msd, see #39
- [x] `tenFHcorr` ->  will be merged into `wlr` given the `rg` @LittleBeannie 
- [x] `tensurv` -> `counting_process` @lili-ling-msd, see #40
- [x] `wMB` -> `mb_weight` @XintongLi2023 , see #41
- [x] `simPWSurv` -> `sim_pw_surv`? @XintongLi2023, see #42
- [x] `simfix` -> `sim_fixed_n`? @XintongLi2023, see #43

**Arguments to be updated:**
- [x] `enrollRates` -> `enroll_rate` @lili-ling-msd, see #44
- [x] `failRates` -> `fail_rate`, @XintongLi2023, see #45
- [x] `failRate` -> `fail_rate` @XintongLi2023, see #45
- [x] `dropoutRates` -> `dropout_rate` @lili-ling-msd, see #46
- [x] `dropoutRate` -> `dropout_rate` @lili-ling-msd, see #46
- [x] `sampleSize` -> `sample_size` @XintongLi2023 , see #47
- [x] `targetEvents` -> `target_event` @lili-ling-msd, see #48
- [x] `timingType` -> `timing_type` @XintongLi2023 , see #49
[...]

PR #28, opened 2023-01-24
update internal variables to follow snake case

LittleBeannie:
I updated most of the internal variables in the snake case. 
I will leave the API for @XintongLi2023 and @lili-ling-msd as described in issue #25.

It is ready for review and merge now.


nanxstats (approved):
I didn't check every specific naming but since this is passing the build I think we should merge. Thanks for taking the hard work!

PR #30, opened 2023-01-26
Re-organize pkgdown reference website

LittleBeannie:
- move old functions to inst
- re-organize pkgdown reference website
- correct a wrong file name


nanxstats (approved):
Great - I fixed a few typos and rephrased the headings in `_pkgdown.yml`.

Not sure if we need the `navbar` field here because the default navigation bar menu items layout is highly similar, so I removed it. If we need to do more detailed customizations for the navbar in the future, we can add it back.

PR #99, opened 2023-05-30
Add workflow vignette

nanxstats:
This PR adds the vignette written by Keaven. It provides a big picture view on data manipulations involved in TTE simulations.

I added the "simulating group sequential trial: one scenario algorithm" workflow diagram to it.


LittleBeannie (approved):
Now we have another valuable vignette for simtrial beginners!

PR #101, opened 2023-05-31
Add reference for datasets

nanxstats:
This PR updates the `@references` field in the datasets previously marked as "TBD" with the proper paper reference.


LittleBeannie (approved):
Thanks for updating these TBD!

PR #103, opened 2023-05-31
delete `tenFHcorr` and merge it into `wlr`

LittleBeannie:


nanxstats (approved):
I applied some style fixes following the tidyverse style guide.

The internal function probably needs a non-placeholder name than `foo`. Also, defining functions within functions (nested functions) is allowed in R, but it's generally not recommended for several reasons: lower performance (each time the outer function is called, the inner function is redefined), confusion in lexical scope and variable shadowing, and decreased maintainability as the inner function is only callable within the outer function.

However, this can be a tomorrow problem, so I will let it pass this time.

Issue #104, opened 2023-06-06
Rename `foo()`


nanxstats:
There is an internal function defined inline named `foo()`:

https://github.com/Merck/simtrial/blob/b4aeaa730465171b295552cb80ad9ad87eb7bdbd/R/wlr.R#L183

This function should be updated:

1. Rename it to use a meaningful function name. See [foo](https://developers.google.com/style/word-list#foo) in Google's developer documentation style guide.
1. Define it outside the export function. This is because currently this is only accessible within the parent function, and it is being redefined again every time the parent function is called, which is suboptimal for performance.

fb-elong:
How about we call it `z_stat`? If it is OK, you can assign it to me :)

nanxstats:
@fb-elong Sounds great! One small detail: how about `wlr_z_stat()`? I think the `wlr_` prefix clearly indicates that the function is related to the weighted logrank test and makes it easier for developers to identify the related functions. This also leaves space for generalization and extension for future testing methods.

nanxstats:
Fixed by #105

PR #105, opened 2023-06-07
rename foo (#104)

elong0527:
Address #104


LittleBeannie (approved):
Thanks for properly addressing it!

nanxstats (changes requested):
Great - I would really appreciate it if you can do me a favor by also addressing the second point in #104: move this function definition outside of `wlr()` for optimal consistency, clarity, and performance.

nanxstats (approved):

PR #108, opened 2023-07-17
update DESCRIPTION

LittleBeannie:


fb-elong (approved):

nanxstats (approved):
Nice, I'm surprised that the dependency is not **actually** used so it can be removed cleanly from `DESCRIPTION`...

PR #110, opened 2023-08-02
Enhancement for doFuture implementation and stable parallel RNG 

### LINKED ISSUE #77: Refactor parallelization code using `%dofuture%`
For simtrial's parallelization implementation, I think the goals should be:

- Use L'Ecuyer-CMRG random seed generation
- Reproducible results
- Parallel backend-independent
- Allow additional levels of nested parallel loops in user code
- Hide the low-level details from users

The tidymodels team handles the L'Ecuyer-CMRG seed generation internally (see `generate_seed()`, tidymodels/tune issue 377, and their [vignette](https://tune.tidymodels.org/articles/extras/optimizations.html#parallel-processing)) - and it works for the mainstream backends including doParallel, doRNG, and doFuture. Understanding their approach first could help us build a robust parallelization solution here.

cmansch:
Resolves #77

- Removes `seed` from the `sim_fixed_n()` function.
- Adds L'Ecuyer-CMRG random seed generation.
- Seed and choice of backend implementation now controlled by the user.
- Backend support prefers future backends (e.g., `plan(multisession)`).


nanxstats (approved):
I think this patch is perfect! With `%dofuture%`, all of this becomes so simple. Thanks so much for looking into this and let's merge it.

If anything can be added to make this even better, that would be a dedicated vignette detailing the parallel recipes - including how to `plan()` properly, demonstrate reproducibility, and demonstrate nested parallel loops across different scenarios. Those can be done in a new PR for another day.

PR #112, opened 2023-08-14
Import `methods::is()` to remove `R CMD check` warning

jdblischak:
Very minor follow-up to PR #110 

The use of `methods::is()` to check the future plan:

https://github.com/Merck/simtrial/blob/341f77f0a598dc6d638bd5c48746952a7db88255/R/sim_fixed_n.R#L273

generated the following warning from `R CMD check`:

```R
Consider adding
  importFrom("methods", "is")
to your NAMESPACE file (and ensure that your DESCRIPTION Imports field
contains 'methods').
```


nanxstats (approved):
Nice. Since the r-lib/actions GHA workflows don't treat notes as errors by default, these things can easily slip through and accumulate. Also, thanks for styling the code 👍

PR #113, opened 2023-09-25
New vignette for sim_fixed_n() in parallel

cmansch:
Adding a new vignette to show users how to set up the backend topologies for parallel runs of sim_fixed_n(). The schema.png is included in the vignette.


nanxstats (approved):
Very well written vignette! I made some minor improvements on code style. I think this is ready for merge.

We may consider two things in the future:

1. Restrict to using maximum 2 cores to ensure [CRAN check compliance](https://cran.r-project.org/web/packages/policies.html).

   GitHub Actions runners [use 2-core or 3-core VMs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) so the pkgdown site should render with proper parallel results.

2. Make the language more generic as we expand the native parallelization availability beyond `sim_n_fixed()`.

Issue #111, opened 2023-08-04
Migrate table backend to data.table


nanxstats:
We will work on rewriting the table backend of simtrial from dplyr/tidyr to data.table to improve simulation performance. Preliminary benchmarks show that a 2x - 10x speedup is possible for a few existing functions driven by different data sizes and table operations involved.

jdblischak:
Intermediate progress update. I've converted `cut_data_by_date()`, `get_cut_date_by_event()`, `counting_process()`, and `sim_fixed_n()` to data.table.

Using simtrial from 341f77f0a598dc6d638bd5c48746952a7db88255, running `sim_fixed_n()` with 100 simulations takes ~22 seconds:

```R
library("simtrial")
library("foreach")
library("future")
plan(sequential)
library("microbenchmark")
microbenchmark(suppressMessages(sim_fixed_n(n_sim = 100)), times = 10)
## Unit: seconds
##                                        expr      min       lq     mean   median       uq      max neval
##  suppressMessages(sim_fixed_n(n_sim = 100)) 20.48683 22.18765 22.42812 22.70253 23.31057 23.51496    10
```

Using simtrial from [5eef141](https://github.com/jdblischak/simtrial/commit/5eef1411cf46d8e9f91459f55da40beeefcedd3f) on my branch ["data.table"](https://github.com/jdblischak/simtrial/commits/data.table), this now takes ~10 seconds:

```R
devtools::load_all(".")
library("foreach")
library("future")
plan(sequential)
library("microbenchmark")
microbenchmark(suppressMessages(sim_fixed_n(n_sim = 100)), times = 10)
## Unit: seconds
##                                        expr      min       lq     mean   median       uq      max neval
##  suppressMessages(sim_fixed_n(n_sim = 100)) 8.311332 8.527891 10.86475 9.451404 12.88969 17.15582    10
```

jdblischak:
I've mostly converted `sim_pw_surv()`. There still remains a single `group_by()`+`summarize()` due to the difficulty in reproducing the internal sorting of `group_by()`. Still made a difference of a few seconds when doing 100 simulations:

```R
microbenchmark(suppressMessages(sim_fixed_n(n_sim = 100)), times = 10)
## Unit: seconds
##                                        expr      min       lq    mean   median       uq      max neval
##  suppressMessages(sim_fixed_n(n_sim = 100)) 7.398563 7.458874 7.57949 7.581341 7.699942 7.762261    10
```

[...]

PR #114, opened 2023-10-05
Convert sim_fixed_n() to use data.table internally

jdblischak:
This PR begins the migration to use data.table as the table backend (#111). I started by converting the function `sim_fixed_n()` (and all its downstream functions).  This resulted in a ~5x speedup (https://github.com/Merck/simtrial/issues/111#issuecomment-1690597131)

While working on the data.table conversion, I also made a few other changes:

* I added the packages stats and utils to the list of Imports. While these are typically always installed alongside the base package, it's best practice to list all the packages your package uses
* I added the trigger `workflow_dispatch` to the `R CMD check` workflow, which allows you to manually trigger a pipeline from any branch. The workflow is setup to only run on the "main" branch, and I like to be able to test the changes I make on a feature branch before opening a PR
* To ensure the changes I made are 100% backwards compatible, I created a test setup where if the environment variable `SIMTRIAL_TEST_BACKWARDS_COMPATIBILITY_REF` is defined, the tests will install that previous version of simtrial, generates results, and then tests that the current version is compatible. This avoids bloating the Git repo and the R package with many `.rds` files
* For the functions that I converted, I made them all return data frames. This ensures the output is compatible with whatever the end user prefers (eg data frame, tibble, data.table). A more sophisticated option would be to capture the class of the input object and then restore that class before returning the object. I didn't do this since it would clutter up the code somewhat, and if no user explicitly requests this feature, it's probably not worth it


nanxstats:
Nice. I fully appreciate the clean approach to compare to the dplyr version for backward compatibility and great context - I think the team will want to see and comment on this.

It would be useful to safeguard these internet-dependent tests using `skip_if_offline()` and `skip_on_cran()` because they can cause trouble in air-gapped environments and CRAN.

nanxstats:
Great discussions today. I guess to ensure the accuracy of our data.table implementation, it's essential to compare the refactored code with the original dplyr implementation. However, there seems to be some potential long-term maintenance concerns if we compare to a specific commit of the original dplyr version, such as:

- Inability to substantially update the new data.table API without causing test failures.
- Risks of the old commit not running properly if dplyr changes its API.

So, I brainstormed two possible solutions to provide flexibility for minor future updates to the dplyr version if needed:

1. Store the dplyr reference implementation in a separate branch: Create a branch, for example, `dplyr-reference-implementation`, store the dplyr version in this branch. Clearly document the purpose of the branch to avoid confusion. Clone and compare to this branch.

2. Dual implementation tests: Maintain the dplyr implementation in the main branch but as live, unit testing code.

What do you think?

nanxstats (approved):
After discussion, the team decides that the plan is to remove the backward compatibility test after finishing the conversion, so we only need to maintain one implementation but not two.

All looking good me. Let's merge this!

PR #115, opened 2023-10-27
Convert mb_weight() to data.table

jdblischak:
I chose `mb_weight()` as the next function to convert since it would allow us to move tidyr from Imports to Suggests.

The speed increase was ~2x, but that wasn't the primary motivation in this case.

```R
library("microbenchmark")

# before
microbenchmark(mb_weight(x, delay = 6, w_max = Inf), times = 1000)
## Unit: milliseconds
##                                  expr     min      lq    mean   median       uq      max neval
##  mb_weight(x, delay = 6, w_max = Inf) 11.0001 11.7249 13.5589 12.19605 13.46255 186.3355  1000

# after
microbenchmark(mb_weight(x, delay = 6, w_max = Inf), times = 1000)
## Unit: milliseconds
##                                  expr    min      lq     mean median      uq      max neval
##  mb_weight(x, delay = 6, w_max = Inf) 3.8054 4.19625 5.571988 4.5562 5.48045 168.4506  1000
```

~~Also note that I had to temporarily disable `skip_if_offline()` because it uses r-project.org, which has been down all morning~~


jdblischak:
I also made a few updates to the vignette that explains `mb_weight()`. I think the section [Freidlin and Korn strong null hypothesis example](https://merck.github.io/simtrial/articles/modestWLRTVignette.html#freidlin-and-korn-strong-null-hypothesis-example) needs some more attention. For example, I don't think this line below about the failure rate is correct:

https://github.com/Merck/simtrial/blob/0f06d961ba41165a7efe2212fd60a066853fdcbf/vignettes/modestWLRTVignette.Rmd#L129

jdblischak:
Also, it would be ideal to add a test case that involves a stratum that gets dropped by the filter:

https://github.com/Merck/simtrial/blob/0f06d961ba41165a7efe2212fd60a066853fdcbf/R/mb_weight.R#L153-L156

nanxstats:
This looks good to me - you will want to check in with the original author of `vignettes/modestWLRTVignette.Rmd` about the minor changes.

nanxstats (approved):

PR #118, opened 2023-10-30
Update pvalue_maxcombo() and move dplyr to Suggests

jdblischak:
As part of #111, I updated `pvalue_maxcombo()`, which allowed me to move dplyr from Imports to Suggests

However, something to discuss, is I also updated the function signature. I found the mysterious `dummy_var` input variable to be confusing. It is only used when called by `group_map()`, and even then it's not obvious from the code how this is happening. Looking at the examples in `?group_map`, they all use the syntax `group_map(~ name_of_function(.x))`, so I updated all the instances of `group_map()` to use this syntax. If `dummy_var` is preferred, I can revert this change.

From a speed perspective, the mean was mostly unchanged, but the median runtime was reduced almost 4x:

```R
# before
library("microbenchmark")
library("simtrial")
x <- sim_fixed_n(
 n_sim = 1,
 timing_type = 5,
 rho_gamma = data.frame(
   rho = c(0, 0, 1),
   gamma = c(0, 1, 1)
 )
)
microbenchmark(pvalue_maxcombo(x), times = 1000)
## Unit: milliseconds
##                expr    min      lq     mean median     uq     max neval
##  pvalue_maxcombo(x) 2.9055 3.75245 5.722706 4.8051 5.6848 34.2989  1000


# after
microbenchmark(pvalue_maxcombo(x), times = 1000, unit = "ms")
## Unit: milliseconds
##                expr    min      lq     mean median      uq     max neval
##  pvalue_maxcombo(x) 0.6253 0.89085 5.068429 1.3186 4.59665 39.6136  1000
```


keaven (approved):

nanxstats (approved):
`dummy_var` also looks a bit mysterious to me... I guess Keaven can tell if you revision makes sense and the intention. I checked the original repo history and this design was already there since the first version in 2019.

PR #119, opened 2023-10-30
Some updates to vignette on Magirr-Burman weights

jdblischak:
Follow up to https://github.com/Merck/simtrial/pull/115#issuecomment-1783230377. After discussing with @keaven, I made a few updates to synchronize the text and code.


nanxstats:
Nice - the code formatter of the `.bib` files came from the "LaTeX Workshop" extension in VS Code, by running its "Sort and align BibTeX file" from the command palette. It's magical.

jdblischak:
> Nice - the code formatter of the .bib files came from the "LaTeX Workshop" extension in VS Code, by running its "Sort and align BibTeX file" from the command palette. It's magical.

Thanks for the tip! That does sound magical

PR #120, opened 2023-10-30
Test mb_weight() with a stratum with no records

jdblischak:
Follow up to https://github.com/Merck/simtrial/pull/115#issuecomment-1783231928. I added a test to confirm the new data.table code can handle the situation where a stratum is dropped because it had no records before the delay ends.

I also confirmed the test fails if I comment out the key line:

https://github.com/Merck/simtrial/blob/7daf1e3270f6f4b0f537358f7260f5871be58c0c/R/mb_weight.R#L148-L149

### DIFF
--- a/tests/testthat/test-double_programming_wMB.R
+++ b/tests/testthat/test-double_programming_wMB.R
@@ -59,3 +59,33 @@ test_that("Validation passed for the situation of multiple stratum", {
   out2 <- data.frame(out2[order(out2$stratum, out2$tte), ])
   testthat::expect_equal(out1, out2)
 })
+
+# Test 3: for the situation where a stratum has no records before delay ends ####
+
+test_that("Validation passed for the situation of a stratum with no records", {
+  set.seed(1)
+  x <- sim_pw_surv(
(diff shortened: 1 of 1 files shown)


tests/testthat/test-double_programming_wMB.R
```diff
@@ -59,3 +59,33 @@ test_that("Validation passed for the situation of multiple stratum", {
...
+      treatment = rep(c("control", "experimental"), 2),
+      duration = rep(1, 4),
+      rate = rep(.001, 4)
+    )
+  ) %>%
+    cut_data_by_event(125) %>%
+    counting_process(arm = "experimental")
+
+  out <- mb_weight(x, delay = 1)
+  testthat::expect_false(anyNA(out$mb_weight))
```
jdblischak, line 90: I'm also open to the opposite:

```suggestion
  testthat::expect_true(!anyNA(out$mb_weight))
```

nanxstats (approved):
Looking good! Thanks again for improving the coverage of the tests. This is also what I call "real tests". 👍

PR #122, opened 2023-11-01
116 develop a new function `get_analysis_date`

LittleBeannie:
This pull request includes the `get_analysis_date` function to enable the calculation of interim/final analysis dates based on various conditions. As it primarily emphasizes the functionality itself, please disregard any concerns regarding code practices (such as data.table/dplyr). It will be forwarded to John for improved coding practices later.


R/get_analysis_date.R
```diff
@@ -0,0 +1,270 @@
...
+#'   min_followup = 12)
+#' get_analysis_date(
+#'   simu_data,
+#'   enroll_rate = enroll_rate,
+#'   min_n_per_stratum = c(200, NA),
+#'   min_followup = 12)
+get_analysis_date <- function(
+    simu_data,
+    # Option 1: planned calendar time for the analysis
+    planned_calendar_time = NA,
```
jdblischak, line 147: It is traditional to specify optional arguments as `NULL`

R/get_analysis_date.R
```diff
@@ -0,0 +1,270 @@
...
+    previous_analysis_date = 0,
+    min_time_after_previous_analysis = NA,
+    # Option 5: minimal follow-up time after specified enrollment fraction
+    enroll_rate = NA,
+    min_n_overall = NA,
+    min_n_per_stratum = NA,
+    min_followup = NA
+){
+  # input checking
+  input_check_scale <- function(x = NA, label = "x"){
```
jdblischak, line 163: Unless there is a compelling reason to do so (eg [function factories](https://adv-r.hadley.nz/function-factories.html)), it is best practice not to nest functions inside of other functions. Nested functions are more difficult to reason about (because of R's [scoping rules](https://adv-r.hadley.nz/environments.html#dynamic-scope)), and more difficult to test and interactively debug (eg you can't run `debug(input_check_scale)`). I recommend moving all helper functions to the bottom of this file.

R/get_analysis_date.R
```diff
@@ -0,0 +1,270 @@
...
+  input_check_scale(min_n_overall, label = "min_n_overall")
+  input_check_scale(min_followup, label = "min_followup")
+
+  input_check_vector <- function(x = NA, label = "x"){
+    if (!(all(is.na(x) | (is.numeric(x) & x > 0)))) {
+      stop(paste0(label, " must be a positive number with either NA or positive numbers!"))
+    }
+  }
+  input_check_vector(target_event_per_stratum)
+  input_check_vector(min_n_per_stratum)
```
jdblischak, line 185: These two are missing the argument `label`

R/get_analysis_date.R
```diff
@@ -0,0 +1,270 @@
...
+  cut_date2 <- pmax(cut_date2a, cut_date2b, na.rm = TRUE)
+
+  # cutting option 3: max time extension to reach targeted events
+  cut_date3 <- max_extension_for_target_event
+
+  # cutting option 4: planned minimum time after the previous analysis
+  cut_date4 <- previous_analysis_date + min_time_after_previous_analysis
+
+  # cutting option 5: minimal follow-up time after specified enrollment fraction
+  get_min_date <- function(enroll_rate, min_n = 400) {
```
jdblischak, line 240: As above, please move this function to the bottom of the file

jdblischak (changes requested):
Great work! I look forward to getting to work on `get_analysis_date()` once it's merged

R/get_analysis_date.R
LittleBeannie: Hello John, I appreciate your 2 valuable suggestions. I chose to use `NA` instead of `NULL` in this scenario to facilitate the implementation of `na.rm = TRUE` while using `pmin()`/`pmax()`. To expedite the development process, I temporarily used `NA`, and once its functionality is approved by Keaven/Yilong, I will hand over the responsibility of refining the code (like `NA` vs. `NULL`, and nested help functions) to you. Thanks!

R/get_analysis_date.R
LittleBeannie: Echo with my reply to the 1st comment.

R/get_analysis_date.R
LittleBeannie: Nice catch! I have just added the 2 labels in the latest commit.

jdblischak (approved):

nanxstats (approved):
I made a few cosmetic improvements here and I think we can merge:

1. Fix typo: prevalence -> prevelance.
1. Rename the first argument as `data` and rename the variable name as `simulated_data`, as `simu_data` doesn't sound familiar.
1. Style code and documentation following tidyverse style guide.
1. Move three named internal functions outside so they won't be refined every time this function is called.

---

I didn't change this but it should be refactored later - these anonymous functions in the two `lapply()` calls have exactly the scoping issue @jdblischak mentioned above:

```r
cut_date5b <- lapply(
  seq_along(min_n_per_stratum),
  function(x) {
    get_min_date(enroll_rate %>% dplyr::filter(stratum == stratum[x]), min_n = min_n_per_stratum[x])
  }
)
```

Since `enroll_rate` is not defined in the input arguments, the function will try to find it in environments by jumping one level above until it is found - you see why this could be dangerous. So having all the inputs properly defined as formal arguments and moving this function as a named function outside is advisable.

PR #127, opened 2023-11-07
124 Rename the function name from `wlr()` to `fh_weight()`

LittleBeannie:

PR #128, opened 2023-11-07
Fix backwards compatibility tests

jdblischak:
I learned from PR #127 (Issue #124) that the backwards compatibility tests aren't currently working as expected. Version 0.3.0 has `wlr()`, not the renamed `fh_weight()`, so `generate-backwards-compatible-test-data.R` should have failed.

The good news is that the backwards compatibility tests are still passing. The problem occurred when I integrated the backwards compatibility tests into the testthat infrastructure, but I had already confirmed they passed before that.


nanxstats (approved):
It has been a mystery to me how exactly namespaces work under testthat... 😅 It's great that you pinpointed this issue and figured out a way to do it properly.

PR #129, opened 2023-11-07
Remove tibble

jdblischak:
This PR removes {tibble} from the package dependencies. Note however that all the functions can still accept {tibble} as input, so all existing code is still valid


R/MBdelayed.R
```diff
@@ -53,7 +53,7 @@
 #'
 #' # MaxCombo with logrank, FH(0,1), FH(1,1)
 #' ten %>%
-#'   fh_weight(rho_gamma = tibble(rho = c(0, 0, 1), gamma = c(0, 1, 1)), return_corr = TRUE) %>%
```
LittleBeannie, line 56: If users input `tibble` rather than `data.frame`, can they still run the codes smoothly?

LittleBeannie (commented):
Hi John, I have 1 question below. Could you please help with it?

R/MBdelayed.R
jdblischak: Correct. The code still runs smoothly if the input is a data frame. Tibbles are enhanced data frames, but ultimately are still data frames, which you can see from their classes, eg

```R
class(tibble::tibble(mtcars))
## [1] "tbl_df"     "tbl"        "data.frame"
```

LittleBeannie: Oh, maybe I am not very clear on my question. Can the code be run if the input is a tibble?

jdblischak: Sorry for the unclear answer. All the functions accept either tibbles or data frames. The output will always be identical.

To be explicit, the code below demonstrates that a tibble or data frame input does not affect the returned result:

```R
identical(
  {set.seed(1); fh_weight(rho_gamma = data.frame(rho = c(0, 0, 1, 1), gamma = c(0, 1, 0, 1)))},
  {set.seed(1); fh_weight(rho_gamma = tibble::tibble(rho = c(0, 0, 1, 1), gamma = c(0, 1, 0, 1)))}
)
## TRUE
```

LittleBeannie: Thanks for explaining this to me!

nanxstats (approved):
This is a small but an important step - appreciate the change!

PR #130, opened 2023-11-09
Skip Breslow's method when there are no ties

jdblischak:
I profiled `sim_fixed_n()` to identify remaining bottlenecks. One I noticed was that Breslow's method is always applied even if there are no ties. Skipping this provides a modest speed increase, which obviously adds up the more simulations you perform.

```R
devtools::load_all()
library("foreach")
library("future")
plan(sequential)
library("microbenchmark")

# before
microbenchmark(suppressMessages(sim_fixed_n(n_sim = 10)), times = 10)
## Unit: milliseconds
##                                       expr      min       lq     mean  median       uq      max neval
##  suppressMessages(sim_fixed_n(n_sim = 10)) 451.3116 492.8935 536.3598 518.449 558.8406 826.8259   100

# after
microbenchmark(suppressMessages(sim_fixed_n(n_sim = 10)), times = 10)
## Unit: milliseconds
##                                       expr      min       lq     mean   median      uq      max neval
##  suppressMessages(sim_fixed_n(n_sim = 10)) 401.6989 434.2817 474.4814 451.7767 478.251 906.1754   100
```

Also, the scenario of ties versus no ties is already well tested

https://github.com/Merck/simtrial/blob/afd9f14050efaac05b66dc52ba3d82ebf7408391/tests/testthat/test-independent_test_counting_process.R#L52

https://github.com/Merck/simtrial/blob/afd9f14050efaac05b66dc52ba3d82ebf7408391/tests/testthat/test-independent_test_counting_process.R#L72


nanxstats (approved):
Saving some unnecessary computations here sounds like a good idea. I'm surprised to see that there is a non-trivial difference. 🤔

PR #132, opened 2023-11-09
123 add a weighting function for early data removal

LittleBeannie:


keaven (approved):

nanxstats (approved):
Looking good! A much welcomed addition... I made some cosmetic improvements.

(There are some new check notes due to not importing dplyr functions since it's now moved to `Suggests` - this is not a problem as it will be automatically resolved after the data.table rewrite.)

PR #133, opened 2023-11-10
Test data.table functionality

jdblischak:
These tests confirm the following:

1. Returned tables always have class "data.frame" (purposefully not using `inherits()` since that would return `TRUE` for tibbles and data tables)
2. The input table is not modified by reference by the data.table commands (thus if in the future someone replaced `as.data.table()` with `setDT()` to increase the speed, these tests would fail)


nanxstats (approved):
Awesome - I think these tests would help simtrial developers write more rigorous code by enforcing consistent behaviors when handling inputs and outputs. 👍

PR #134, opened 2023-11-13
Update parallel.Rmd to align with CRAN Policies

cmansch:
Changing the number of workers used by the parallel example to 2 workers. This should align with the [CRAN Policy](https://cran.r-project.org/web/packages/policies.html).

> "If running a package uses multiple threads/cores it must never use more than two simultaneously: the check farm is a shared resource and will typically be running many checks simultaneously."


LittleBeannie (approved):
Thanks for the updates!

jdblischak (approved):

nanxstats (approved):
I'm glad you found this. 👍 There are many rules in the policy. This is definitely an enforceable one.

PR #136, opened 2023-11-15
Test get_analysis_date()

jdblischak:
I converted the examples to tests. From running covr locally, `get_analysis_date()` now has 95.45% coverage

### DIFF
--- a/tests/testthat/test-get_analysis_date.R
+++ b/tests/testthat/test-get_analysis_date.R
@@ -1,0 +1,124 @@
+# Simulate trial data
+alpha <- 0.025
+ratio <- 3
+n <- 500
+info_frac <- c(0.7, 1)
+prevalence_ratio <- c(0.4, 0.6)
+study_duration <- 48
+# Two strata
+stratum <- c("Biomarker-positive", "Biomarker-negative")
+prevalence_ratio <- c(0.6, 0.4)
+# enrollment rate
+enroll_rate <- gsDesign2::define_enroll_rate(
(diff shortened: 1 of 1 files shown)


jdblischak:
https://app.codecov.io/gh/Merck/simtrial/blob/test-get_analysis_date/R%2Fget_analysis_date.R

nanxstats (approved):
Nice 👍

tests/testthat/test-get_analysis_date.R
```diff
@@ -0,0 +1,124 @@
...
+)
+
+test_that("planned_calendar_time", {
+  observed <- get_analysis_date(simulated_data, planned_calendar_time = 24)
+  expect_equal(observed, 24)
+})
+
+test_that("target_event_overall", {
+  observed <- get_analysis_date(simulated_data, target_event_overall = 300)
+  expect_equal(observed, 25.6150577826303)
```
LittleBeannie, line 51: Could you kindly explain the purpose of the testing done after line 49? The output 25.6150577826303 is an unvalidated result. I understand that you may have performed these tests to verify the data.table transformation. If that's the case, would it be possible to rename the file to something like "test-compare with original function output - get_analysis_date"?

jdblischak: Yes, I need tests in order to refactor the code. This is the only way we will know it continues to behave as you originally designed it. How about `test-unvalidated-get_analysis_date.R`?

PR #137, opened 2023-11-15
Convert `get_analysis_date()` to data.table

jdblischak:
In addition to converting to data.table, I made the following changes:

1. Replaced the vectorized logical operators `&` and `|` with their scalar equivalents `&&` and `||` for use in `if` statements
2. Replaced `lapply()` + anonymous functions with `for` loops
3. Styled the package files

The speed improvement is as expected:

```R
library("simtrial")
library("microbenchmark")
n <- 500
x <- sim_pw_surv(n = n)
microbenchmark(
  get_analysis_date = get_analysis_date(
    x,
    min_n_overall = n * 0.8,
    min_n_per_stratum = c(200, NA),
    min_followup = 12
  )
)

# before

## Unit: milliseconds
##               expr    min     lq     mean median     uq     max neval
##  get_analysis_date 6.7069 7.9107 11.24213 8.5425 9.4881 209.134   100

# after

## Unit: milliseconds
##               expr    min     lq     mean median      uq     max neval
##  get_analysis_date 1.8519 2.1588 2.557967 2.4057 2.63095 10.2379   100
```

### DIFF
--- a/R/get_analysis_date.R
+++ b/R/get_analysis_date.R
@@ -41,6 +41,8 @@
 #'
 #' @return A numerical value of the analysis date.
 #'
+#' @importFrom data.table as.data.table .N
+#'
 #' @export
 #'
 #' @examples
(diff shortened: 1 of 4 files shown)


R/get_analysis_date.R
```diff
@@ -231,16 +235,14 @@ get_analysis_date <- function(
...
-      function(x) {
-        get_cut_date_by_event(data %>% dplyr::filter(stratum == stratum[x]),
-          event = target_event_per_stratum[x]
-        )
-      }
-    ) %>%
-      unlist() %>%
-      max()
+    cut_date2b <- vector(mode = "list", length = length(stratum))
+    for (i in seq_along(target_event_per_stratum)) {
```
LittleBeannie, line 239: Hi John, sorry for the not timely review, I was on a flight last Tuesday and Wednesday. May I know if the change from `lappy` to `for`-loop, is because `lappy` is slow?

jdblischak: No, I didn't remove `lapply()` for speed reasons. `lapply()` is perfectly fast.

The main motivation was scoping (as @nanxstats described in https://github.com/Merck/simtrial/pull/122#pullrequestreview-1711786166. The anonymous function only had a single argument `x`, but it accessed multiple variables outside its scope (`data`, `stratum`, `target_event_per_stratum`). One option would have been to create a new internal function with all the required arguments. But since the code is so minimal (filtering followed by passing to `get_cut_date_by_event()`), I instead decided to put in a `for` loop (which doesn't have the scoping issues that an anonymous function has).

PR #138, opened 2023-11-15
Test input checking functions

jdblischak:
Follow up to #136. I had forgotten to test the input checking functions.

Changes include:

* Renamed `input_check_scale()` to `input_check_scalar()` since it requires a single value
* Fixed a bug in `get_analysis_date()` (both `cond1` and `cond2` checked `min_n_overall`)

My assumptions (please correct me if I'm wrong):

* `input_check_scalar()` requires non-negative numbers (ie accepts zero)
* `input_check_vector()` requires positive numbers (ie rejects zero)
* `min_n_overall` and `sum(min_n_per_stratum)` can equal the sample size (the code and the error message disagreed, and I went with the code)

Open questions:

* Do the input checking functions require whole numbers? Currently they accept floating point numbers
* Should `input_check_vector()` accept a vector of multiple missing values? (`input_check_vector(NA, NA)`)

### DIFF
--- a/R/get_analysis_date.R
+++ b/R/get_analysis_date.R
@@ -190,19 +190,19 @@ get_analysis_date <- function(
     min_n_overall = NA,
     min_n_per_stratum = NA,
     min_followup = NA) {
-  input_check_scale(planned_calendar_time, label = "planned_calendar_time")
-  input_check_scale(target_event_overall, label = "target_event_overall")
-  input_check_scale(max_extension_for_target_event, label = "max_extension_for_target_event")
-  input_check_scale(min_time_after_previous_analysis, label = "min_time_after_previous_analysis")
-  input_check_scale(min_n_overall, label = "min_n_overall")
-  input_check_scale(min_followup, label = "min_followup")
+  input_check_scalar(planned_calendar_time, label = "planned_calendar_time")
+  input_check_scalar(target_event_overall, label = "target_event_overall")
+  input_check_scalar(max_extension_for_target_event, label = "max_extension_for_target_event")
+  input_check_scalar(min_time_after_previous_analysis, label = "min_time_after_previous_analysis")
+  input_check_scalar(min_n_overall, label = "min_n_overall")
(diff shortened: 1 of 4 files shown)


LittleBeannie (approved):
Thanks for correcting my bad wording!

jdblischak:
Yujie answered my questions on Teams. My summary below:

> * Do the input checking functions require whole numbers? Currently they accept floating point numbers

The following arguments require whole numbers: target_event_overall, min_n_overall, target_event_per_stratum, min_n_per_stratum

> * Should `input_check_vector()` accept a vector of multiple missing values? (`input_check_vector(NA, NA)`)

Yes. Directly from Yujie:

> For example, min_n_per_stratum = c(NA, NA, NA) is equivalent to min_n_per_stratum = NA

jdblischak:
Ready to go as long as CI passes

nanxstats (commented):
I wonder if we should use the is whole number implementation directly from `?base::is.integer` code examples.

```r
is_whole_number <- function(x) trunc(x) == x
is.wholenumber <- function(x, tol = .Machine$double.eps^0.5) abs(x - round(x)) < tol

is_whole_number(3.000000000000001)
# [1] FALSE
is_whole_number(3.0000000000000001)
# [1] TRUE

is.wholenumber(3.0000001)
# [1] FALSE
is.wholenumber(3.00000001)
# [1] TRUE
```

The main difference is the tolerance: `.Machine$double.eps` (inexplicitly 1e-16) vs. `.Machine$double.eps^0.5` (explictly 1e-8).

If we are dealing with data that's likely to have minute computational inaccuracies from arithmetic operations, the `is.wholenumber()` method looks safer. Also, it was used in [`gsDesign::toInteger()`](https://github.com/keaven/gsDesign/blob/6de6583d621bf2fd0bff8e4b4f4b755d77ee407d/R/toInteger.R#L159).

nanxstats (approved):
Thanks for the update - looks great

PR #144, opened 2023-11-21
Convert early_zero_weight() to data.table

jdblischak:
Please merge PR #143 first. I will likely need to resolve some merge conflicts locally before we can merge this one

I also added some more tests

The speed increase for the unstratified case was limited (~1200 -> 800 microseconds). However, the speed increase for the stratified case was impressive (~123 -> ~5 milliseconds)

Some lessons learned:

* For a single column, `length(unique(x$col))` is noticeably faster than `uniqueDT(x[, .(col)])`, as least for small tables (nrow = 125 in this case). I assume that `uniqueDT()` pays off when the number of rows increases, and also when multiple columns need to be selected, but I didn't investigate thoroughly
* `merge.data.table(x, y, by = "col", all.x = TRUE)` matches the sort order of `dplyr::left_join(x, y)`, whereas `x[y, on = "col"]` returns a different row order

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: simtrial
 Type: Package
 Title: Clinical Trial Simulation
-Version: 0.3.0.5
+Version: 0.3.0.6
 Authors@R: c(
     person("Keaven", "Anderson", email = "keaven_anderson@merck.com", role = c("aut")),
     person("Yilong", "Zhang", email = "elong0527@gmail.com", role =  c("aut")),
--- a/R/early_zero_weight.R
+++ b/R/early_zero_weight.R
@@ -26,6 +26,8 @@
 #' @return A data frame. The column `weight` contains the weights for the
 #'   early zero weighted logrank test for the data in `x`.
 #'
+#' @importFrom data.table ":=" as.data.table merge.data.table setDF
+#'
 #' @export
 #'
 #' @references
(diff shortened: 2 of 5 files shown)


tests/testthat/test-data.table.R
```diff
@@ -69,6 +75,51 @@ test_that("functions that use data.table do not modify input data table", {
...
+  x_original <- data.table::copy(x)
+  early_zero_weight(x, early_period = 2)
+  expect_identical(x, x_original)
+  # stratified
+  x <- data.frame(
+    stratum = rep(c("Biomarker-negative", "Biomarker-positive"), c(3L, 6L)),
+    events = rep(1, 9L),
+    n_event_tol = c(0, 1, 1, 0, 1, 0, 1, 1, 0),
+    tte = c(
+      1.5993802300052806, 2.0909139412967335, 10.506616675248615,
```
nanxstats, line 92: After a while, people would wonder where all these magic numbers came from - I'd put the code to generate them here for clarity.

jdblischak: The code to produce the input data is 40+ lines of code. I put the full code to produce it in the `early_zero_weight()` tests. However, these data.table tests are simply to confirm the data.table behavior is working correctly, so I try to minimize the amount of extraneous setup code.

jdblischak: How about I add a comment stating that this data was created using the code that is available from the example in the function documentation?

jdblischak: Actually I don't care that strongly about this. I went ahead and duplicated the code

nanxstats (approved):
Thanks for rewriting this, looking good to me.

Also, thanks for removing the magic numbers in tests - i think this will pay off in the long run.

PR #146, opened 2023-11-27
Use base pipe operator

### LINKED ISSUE #145: Replace magrittr pipe with native pipe and remove dependency
Given these facts:

- The native pipe has been available for 2.5 years.
- We were only using the basic `%>%` operator.
- There might be a slight performance benefit of using the base pipe.
- We want to minimize the footprint of hard dependencies.
- Base pipes may gain special aesthetics when using fonts with programming ligatures (such as Cascadia Code, Fira Code, Hasklig).

We may want to consider replacing the `%>%` pipes with the native pipe `|>` and remove the magrittr dependency.

nanxstats:
Fixes #145 by replacing the magrittr pipe `%>%` with the native pipe operator `|>`.

- As a natural result, now it uses `Depends: R (>= 4.1.0)`.
- Also fixes the `R CMD check` note on should not start the `Description` field with the package name or "This package" in `DESCRIPTION`.
- Also fixes a typo on "prevalence".


jdblischak (approved):

jdblischak:
I realized there is a potential negative consequence of this change. Now that the magrittr pipe is no longer exported in `NAMESPACE`, a user's existing code may fail. For example, this will no longer work:

```R
library(simtrial)
x <- sim_pw_surv(n = 400)
y <- cut_data_by_event(x, 150) %>% counting_process(arm = "experimental")
## Error in cut_data_by_event(x, 150) %>% counting_process(arm = "experimental") : 
##   could not find function "%>%"
```

To be clear, I still think we made the right decision. This package is pre-CRAN release, so it's ok for us to make some breaking changes. Also, presumably most users that use the pipe will also load dplyr, so the magrittr pipe would still be exported. And if a user does get this error message, now they can search and find this explanation.

PR #150, opened 2023-11-27
Cleanup ignore files

nanxstats:
This PR cleans up `.gitignore` and `.Rbuildignore` by removing the superfluous rules and reordering the lists.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -1,12 +1,11 @@
-^\.github$
 ^.*\.Rproj$
 ^\.Rproj\.user$
-^LICENSES_THIRD_PARTY$
-^_pkgdown\.yml$
 ^\.gitattributes$
+^\.github$
+^codecov\.yml$
+^_pkgdown\.yml$
 ^pkgdown$
+^docs$
+^data-raw$
 ^LICENSE$
 ^LICENSES_THIRD_PARTY$
-^codecov\.yml$
-^doc$
-^data-raw$
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,6 @@
 .Rproj.user
-doc
-docs
-Meta
-inst/doc
-.Rprofile
 .Rhistory
 .RData
 .Ruserdata
-tests/testthat/fixtures/backwards-compatibility
+inst/doc
+docs


jdblischak:
Are you sure that `doc` and `Meta` are no longer sometimes created when regenerating the package documentation? I couldn't find anything about this in either the [roxygen2 NEWS](https://github.com/r-lib/roxygen2/blob/main/NEWS.md) or the [devtools NEWS](https://github.com/r-lib/devtools/blob/main/NEWS.md). In fact, both roxygen2 and devtools list `doc` and `Meta` in both their `.gitignore` and `.Rbuildignore` files. These might just be stale, but I can't find any assurance that the problem of these directories has been resolved.

nanxstats:
> Are you sure that `doc` and `Meta` are no longer sometimes created when regenerating the package documentation? I couldn't find anything about this in either the [roxygen2 NEWS](https://github.com/r-lib/roxygen2/blob/main/NEWS.md) or the [devtools NEWS](https://github.com/r-lib/devtools/blob/main/NEWS.md). In fact, both roxygen2 and devtools list `doc` and `Meta` in both their `.gitignore` and `.Rbuildignore` files. These might just be stale, but I can't find any assurance that the problem of these directories has been resolved.

I'm not aware of a single case where these directories are created under the source package from my past experience, that's why I think these rules should be removed. I wonder if the original template author @elong0527 can provide a positive example. If not, we should remove them.

Apparently, `Meta/` and `doc/` are directories that appear in the binary package - see the [R Packages book](https://r-pkgs.org/structure.html#fig-package-files). I'd say it's highly usual if they appear in the source package, and I'm curious what workflow could create them under the source package.

jdblischak:
> I'm not aware of a single case where these directories are created under the source package from my past experience, that's why I think these rules should be removed.

We can remove them. Worst-case scenario we add them back

> I'd say it's highly unusual if they appear in the source package, and I'm curious what workflow could create them under the source package.

It used to happen to my packages all the time when using roxygen2/devtools/pkgdown to document and build the package. It may have been fixed (I wouldn't have noticed since I started ignoring them), but I couldn't find any documentation for this. There are 6k+ GitHub repos that include `Meta` and `doc` in their `.Rbuildignore`, so at least at one point it was a very common problem

https://github.com/search?q=path%3A.Rbuildignore%20Meta&type=code

nanxstats:
Found it: `devtools::build_vignettes()` will create `doc/` and `Meta/` without cleaning up. I'm not sure if this is a popular tool, but I'm ok about adding them back. To be honest, I just click the Knit button and it just works...

jdblischak:
> Found it: `devtools::build_vignettes()` will create `doc/` and `Meta/` without cleaning up.

Thanks for finding it! I was frustrated that I couldn't find the source. I thought it was either `devtools::document()` or `pkgdown::build_site()`. But I do call `devtools::build_vignettes()`  in my workflowr convenience script [`document.R`](https://github.com/workflowr/workflowr/blob/79a96d49b27187f9353581fda6ff6d900c2bb42c/scripts/document.R), which explains why I manually delete `Meta` and `doc`

> I'm not sure if this is a popular tool, but I'm ok about adding them back. To be honest, I just click the Knit button and it just works...

Let's not worry about it for now. If it becomes a nuisance in the future, we can add it back then

nanxstats:
Wow, I guess if devtools has a designated output directory like `R CMD check` does (.Rcheck), things like this would be less of a problem...

PR #174, opened 2023-12-05
Use inverse CDF method C++ implementation in `rpwexp()`

nanxstats:
This PR lets `rpwexp()` use the inverse CDF C++ implementation. More specifically:

- The underlying implementation for `rpwexp()` now uses the inverse CDF method C++ version for optimal speed.
- The original `rpwexp()` (R implementation of the naive method) is now renamed and archived as an internal function.
- Previously exported C++ functions for the naive and inverse CDF methods are un-exported to minimize the namespace.

In case one needs to access the legacy functions, here is the **mapping of before and after**:

- `rpwexpRcppInv()` is now removed. Use `rpwexp()` to use it directly.
- `rpwexp()` is now `simtrial:::rpwexp_naive()`.
- `rpwexpRcpp()` is now `simtrial:::rpwexp_naive_cpp()`.

Benchmark showing things are working as expected after the change:

```r
microbenchmark::microbenchmark(
  rpwexp(
    n = 1e4,
    fail_rate = data.frame(rate = c(1, 3, 10), duration = c(.5, .5, 1))
  ),
  simtrial:::rpwexp_naive(
    n = 1e4,
    fail_rate = data.frame(rate = c(1, 3, 10), duration = c(.5, .5, 1))
  ),
  simtrial:::rpwexp_naive_cpp(
    n = 1e4,
    fail_rate = data.frame(rate = c(1, 3, 10), duration = c(.5, .5, 1))
  ),
  times = 5000
)
#>     min       lq     mean  median       uq       max neval cld
#> 152.889 166.0090 182.4453 169.125 173.3890  4478.184  5000   a
#> 671.375 745.9335 819.5745 768.504 788.4505 75598.916  5000   b
#> 650.383 720.4315 757.5569 736.606 752.4525  4926.970  5000   c
```


keaven (approved):
Very nice! Easier for users while maintaining legacy routines unexported.

jdblischak (approved):

jdblischak:
I searched for all instances of `rpwexp()`. The only thing I noticed was that the archived function `simPWSurv_()` also calls `rpwexp()`:

https://github.com/Merck/simtrial/blob/db7bae32185e1c89c41496466725d8560b46e203/inst/archive/simPWSurv_.R#L125-L126

If the hope is that these functions can continue to run as previously, I think these should be replaced with `rpwexp_naive()`. On the other hand, if these archived functions are purely for reading purposes, then the bit rot is ok.

PR #175, opened 2023-12-06
Update news for simtrial 0.3.1 release

nanxstats:
This PR updates `NEWS.md` for simtrial 0.3.1 release.

Also extends the installation section to include CRAN installation guide in `README.md`.

Besides, I fixed a few minor but potential issues per [extrachecks](https://github.com/DavisVaughan/extrachecks), such as

- Add a reference to the description field in `DESCRIPTION`.
- Avoid using `:::` in tests.
- Update copyright year in `LICENSE`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -25,9 +25,9 @@ Description: Provides some basic routines for simulating a
     generate trial simulations for trials with time to event outcomes.
     Piecewise exponential failure rates and piecewise constant
     enrollment rates are the underlying mechanism used to simulate
-    a broad range of scenarios. However, the basic generation of data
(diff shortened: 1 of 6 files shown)


jdblischak (approved):

jdblischak:
> Avoid using `:::` in tests.

Out of curiosity, could you please expand on the motivation for this? I didn't see it in extrachecks. The main advantage I see of using `:::` is that you can copy-paste the code to interactively test it even when you only have package loaded the standard way (ie with `library()` and not `devtools::load_all()`). The only downside I see if that prefixing with `simtrial:::` is that it is more verbose. Are there more potential downsides than verbosity?

keaven (approved):

PR #181, opened 2023-12-11
Add CRAN badges with proper links

nanxstats:
This PR updates the badges in `README.md` to have proper links now that the package is on CRAN.


jdblischak (approved):
:tada:

PR #186, opened 2024-01-04
Enrich documentation of `get_analysis_date()`

### LINKED ISSUE #140: Vignette for `get_analysis_date()`
Need a vignette for `get_analysis_date()`.

LittleBeannie:
Fixes #140


R/get_analysis_date.R
```diff
@@ -16,7 +16,25 @@
 #  You should have received a copy of the GNU General Public License
 #  along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-#' Get the analysis date under multiple conditions
+
+#' Derive analysis date for interim/final analysis given multiple conditions.
+#'
+#' To obtain the analysis date, consider the following multiple conditions:
+#' Condition 1: The planned calendar time for analysis.
```
jdblischak, line 23: This will not be nicely formatted in the final help file. This is what it looks like locally:

image

jdblischak: I recommend either using `\describe{}`, which looks like the following:

```
#' To obtain the analysis date, consider the following multiple conditions:
#' \describe{
#'   \item{Condition 1:}{The planned calendar time for analysis.}
#'   \item{Condition 2:}{The targeted events, encompassing both overall population
#' and stratum-specific events.}
#' }
```

which would result in

image

or a Markdown list

```
#' * Condition 3: The maximum time extension required to achieve the targeted events.
#' * Condition 4: The planned minimum time interval after the previous analysis.
#' * Condition 5: The minimum follow-up time needed to reach a certain number of
#' patients in enrollments.
```

which would render as

image

jdblischak (changes requested):

R/get_analysis_date.R
LittleBeannie: Hi @jdblischak , thank you for reviewing the formatting! I have just made an additional commit. Please feel free to make any edits or suggest improvements to the content if you think there are any further areas that could benefit from refinement.

R/get_analysis_date.R
```diff
@@ -19,14 +19,14 @@
 
 #' Derive analysis date for interim/final analysis given multiple conditions.
 #'
-#' To obtain the analysis date, consider the following multiple conditions:
```
jdblischak: I liked this introductory line. It's a bit jarring to get dropped directly into the list of multiple conditions

R/get_analysis_date.R
```diff
@@ -35,7 +35,7 @@
 #' where date1, date2, date3, date4, date5 represent the analysis dates determined
 #' solely by Condition 1, Condition 2, Condition 3, Condition 4 and Condition 5,
 #' respectively.
-#'
+#' }
```
jdblischak, line 38: This closing bracket should be after the final `\item{}{}` line. Only text within `\item{}{}` should be contained with the `\describe{}`

jdblischak (changes requested):

R/get_analysis_date.R
nanxstats: I agree. This line is now restored before the list.

R/get_analysis_date.R
nanxstats: Good point

nanxstats (approved):
Nice additions. I refined the description list content and format following the suggestions from @jdblischak, made it a `@details` section, and moved it down.

If something needs 8 examples in 230 lines of code and extensive code comments to explain, I highly recommend writing a vignette instead of overutilizing the function documentation. Vignettes simply has much better support for long format documentation and makes more sense in this case. But since this is your package, I will respect your choice.

R/get_analysis_date.R
nanxstats: Great suggestions

nanxstats:
GitHub will let me merge once @jdblischak approves the requested changes.

jdblischak (approved):

PR #187, opened 2024-01-25
Cleanup CI files

jdblischak:
Follow up to https://github.com/Merck/gsDesign2/pull/326. Nan already set the default token to be read-only

Made the following minor changes:

* Added a manual trigger via `workflow_dispatch` to run CI on feature branches as needed
* Removed unnecessary `GITHUB_PAT` env vars (this would only be useful if the workflows were installing packages from GitHub)
* Simplified the branches filter to the main branch
* Only push the updated pkgdown site to the gh-pages branch when running on the main repo (ie not forks)


nanxstats (approved):
Looking good, thanks for unifying the workflows across repos.

PR #195, opened 2024-02-09
Implement `sim_gs_n()`

### LINKED ISSUE #194: Develop `sim_gs_n` for group sequential design simulations
@LittleBeannie will upload her initial draft on functional input for cutting, and @jdblischak will improve the syntax via the functional factory.

LittleBeannie:
Closes #194


jdblischak:
>  Given the numerous updates in this PR, would you prefer to merge it first and then create a PR specifically for the functional factory?

I like that approach

R/wlr.R
```diff
@@ -0,0 +1,66 @@
...
+  if ("fh" %in% class(weight)) {
+    ans <- data |>
+      counting_process(arm = "experimental") |>
+      fh_weight(rho_gamma = data.frame(rho = weight$rho, gamma = weight$gamma))
+
+  } else if ("mb" %in% class(weight)) {
+    ans <- data |>
+      counting_process(arm = "experimental") |>
+      mb_weight(delay = weight$delay, w_max = weight$w_max) |>
+      dplyr::summarize(
```
jdblischak, line 50: I can also convert this to {data.table} since we have moved {dplyr} to Suggests

jdblischak (approved):

nanxstats (approved):
Great, let's merge this first to move it forward. I believe @jdblischak can show us the ideal functional interface and a performant data.table backend. 🥇 

Besides the important things above... it would be great if you could also include a few minor improvements on code style:

- Avoid adding rlang back to `Imports`, considering we have just made an effort to remove it from there.
- `1:n` should be `seq_len(n)`.
- `"zzz" %in% class(x)` should be `inherits(x, "zzz")`.
- Use year 2024 in the copyright header.
- Boilerplate code like this:

  ```r
  ans <- list(rho = rho, gamma = gamma)
  class(ans) <- c(class(ans), "fh", "wlr")
  return(ans)
  ```

  can be simplified to a one-liner:

  ```r
  structure(list(rho = rho, gamma = gamma), class = c("list", "fh", "wlr"))
  ```

  See [r-base-shortcuts](https://github.com/nanxstats/r-base-shortcuts#create-and-assigning-s3-classes-in-one-step).
- It's a bit unclear why we should have `name()` and `name_weight()` functions coexist where `name` = `fh`, `mb`, `early_zero`,... Perhaps this duplication will become obsolete after the API refactor - having a single source of truth and not having to read documentation to use functions correctly would be great.

PR #198, opened 2024-02-16
Refactor `do.call(rbind)` with `data.table::rbindlist()` in RMST

### LINKED ISSUE #193: Refactor `do.call(rbind)` with `data.table::rbindlist()` in RMST
There is some `do.call(rbind)` usage in `R/rmst.R`.

Since we're already building on data.table, it would be great to use `data.table::rbindlist()`to handle such patterns instead (while still returning data.frame).

This might not be an obvious issue but is an obvious opportunity to safeguard from a performance perspective, to make sure it won't be a potential bottleneck if the list size grows large (see [Jim Hester's benchmarking article](https://rpubs.com/jimhester/rbind) on this topic).

jdblischak:
Closes #193

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: simtrial
 Type: Package
 Title: Clinical Trial Simulation
-Version: 0.3.2.2
+Version: 0.3.2.3
 Authors@R: c(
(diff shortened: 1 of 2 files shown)


LittleBeannie (approved):
Thanks for the updates, @jdblischak !

nanxstats (approved):
Thanks for keeping us in line with best practices!

PR #200, opened 2024-02-21
Fix some NOTEs from R CMD check

jdblischak:
I fixed the 3 current NOTEs from `R CMD check`. Below is the output from when I ran it locally. They are also viewable from the the [most recent build log](https://github.com/Merck/simtrial/actions/runs/7935871469/job/21669939604#step:6:145).

```
── R CMD check results ───────────────────────────────── simtrial 0.3.2.3 ────
Duration: 2m 41.1s

❯ checking R code for possible problems ... NOTE
  wlr: no visible binding for global variable 'v'
  wlr: no visible binding for global variable 'z'
    v z

❯ checking Rd line widths ... NOTE
  Rd file 'sim_gs_n.Rd':
    \examples lines wider than 100 characters:
       # - At least 20 months have elapsed after enrolling 200/400 subjects, with a minimum of 20 months follow-up

  These lines will be truncated in the PDF manual.

❯ checking for unstated dependencies in vignettes ... NOTE
  'library' or 'require' call not declared from: 'gt'

0 errors ✔ | 0 warnings ✔ | 3 notes ✖

R CMD check succeeded
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: simtrial
 Type: Package
 Title: Clinical Trial Simulation
-Version: 0.3.2.3
+Version: 0.3.2.4
 Authors@R: c(
     person("Keaven", "Anderson", email = "keaven_anderson@merck.com", role = c("aut")),
     person("Yilong", "Zhang", email = "elong0527@gmail.com", role =  c("aut")),
@@ -56,6 +56,7 @@ Suggests:
     ggplot2,
     gsDesign,
     gsDesign2,
+    gt,
     knitr,
     rmarkdown,
     survMisc,
--- a/R/global.R
+++ b/R/global.R
@@ -55,8 +55,10 @@ utils::globalVariables(
     "time",
     "treatment",
     "tte",
+    "v",
     "var_o_minus_e",
-    "weight"
+    "weight",
+    "z"
   )
 )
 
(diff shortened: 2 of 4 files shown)


nanxstats (approved):
Awesome! We should keep an eye on the build status regularly, otherwise, it will become more difficult to fix after all the minor issues accumulate.

jdblischak:
> We should keep an eye on the build status regularly, otherwise, it will become more difficult to fix after all the minor issues accumulate.

If it becomes a concern, we can configure `error_on` to fail if `R CMD check` produces a `NOTE`. This would make our check much more stringent, but it would also prevent these NOTEs from accumulating.

https://github.com/r-lib/actions/tree/v2-branch/check-r-package#usage

nanxstats:
> > We should keep an eye on the build status regularly, otherwise, it will become more difficult to fix after all the minor issues accumulate.
> 
> If it becomes a concern, we can configure `error_on` to fail if `R CMD check` produces a `NOTE`. This would make our check much more stringent, but it would also prevent these NOTEs from accumulating.
> 
> https://github.com/r-lib/actions/tree/v2-branch/check-r-package#usage

Solid advice! 👍  At the moment, we can keep the default behavior. Given how frequently using tidy evaluation creates those "no visible binding for global variable" notes, things could derail from "stringent" to "annoying" quickly... 😂 If we release frequently enough, the check notes situation should be generally manageable.

PR #201, opened 2024-02-22
Use function factory pattern to pass cutting functions to `sim_gs_n()`

### LINKED ISSUE #196: Refactor `sim_gs_n()` to improve interface and performance
It would be great to leverage the functional programming patterns available in R, such as [function factories](https://adv-r.hadley.nz/function-factories.html), to avoid the explicit quoting and evaluation in the current implementation.

The goal is to enable users to use the established functions and extend them freely following a specification.

This also involves data.table rewrite for the current dplyr implementation.

This is a follow up to https://github.com/Merck/simtrial/pull/195.

jdblischak:
Follow-up to #195. Closes #196.

This PR makes the following updates:

* Instead of passing a list of quoted `get_analysis_date()` expressions that then need to be evaluated, the argument `cutting` of `sim_gs_n()` now accepts a list of cutting functions. These functions are created by the function factory `create_cutting()`
* Instead of passing quoted test functions to the argument `test` of `sim_gs_n()`, only the test function itself is passed, and any further arguments are passed to the testing function via `...`
* Various other code improvements

Note that I didn't fully convert `maxcombo()`. I don't think I fully understand the purpose of this function. Its arguments are quoted functions, but they are never evaluated. Instead their arguments for `rho` and `gamma` are extracted and passed to `fh_weight()`. Is there any reason that `maxcombo()` couldn't instead just receive the vectors of `rho` and `gamma` directly? And why is `maxcombo()` needed when `fh_weight()` already exists?


LittleBeannie:
Thanks for applying the function factory, @jdblischak ! Could you please introduce us to the changes and new syntax tomorrow?

nanxstats:
Overall looking good. Let's try follow the testing best practices in [designing your test suite](https://r-pkgs.org/testing-design.html):

- Avoid the `library(gsDesign2)` call and qualify namespaces explicitly.
- Move the logic at top of the test file with "test file scope" into a function within a "helper" file and call the function to setup things at the start of individual tests.

jdblischak:
> Before the merge, could you please introduce us to the changes and new syntax tomorrow?

@LittleBeannie Yes, of course. I'll explain everything in today's meeting. Also note that this morning I also converted the `maxcombo()` interface to accept vectors instead of quoted function expressions.

LittleBeannie (approved):
Thanks for these terrific improvement, @jdblischak ! I will add an example of maxcombo test after this PR is merged.

nanxstats (approved):
Awesome! I love simple, reliable, and effective solutions like this.

PR #206, opened 2024-02-25
Make tests self-contained and suppress printing

nanxstats:
As recommended by [R Packages](https://r-pkgs.org/testing-design.html), this PR makes all existing tests self-contained by creating helper functions to wrap the top-level logic that used to have awkward "test file scope". Now all test files only contain individual `test_that()` blocks.

I also used `invisible(capture.output())` to suppress printing from tests for MaxCombo and WLR, where `survMisc::comp()` calls will generate long printed outputs.


LittleBeannie (approved):
Thanks, @nanxstats!

PR #208, opened 2024-02-27
add sample size and events to `sim_gs_n` per analysis per simulation

### LINKED ISSUE #207: More output of `sim_gs_n`
It would be good for `sim_gs_n` to return sample size (n) and events (event) for each record returned.

LittleBeannie:
Closes #207

### DIFF
--- a/R/sim_gs_n.R
+++ b/R/sim_gs_n.R
@@ -227,6 +227,8 @@ sim_gs_n <- function(
       ans_1sim_new$analysis <- i_analysis
       ans_1sim_new$cut_date <- cut_date[i_analysis]
       ans_1sim_new$sim_id <- sim_id
+      ans_1sim_new$n <- nrow(simu_data_cut)
+      ans_1sim_new$event <- sum(simu_data_cut$event)
 
       # rbind simulation results for all IA(s) and FA in 1 simulation
       ans_1sim <- rbind(ans_1sim, ans_1sim_new)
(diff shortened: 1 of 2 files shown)


nanxstats (changes requested):
Instead of removing `n` and `event` from `observed`, I think the better way to update the tests is to add these to `expected`.

LittleBeannie:
> Instead of removing `n` and `event` from `observed`, I think the better way to update the tests is to add these to `expected`.

Agreed! I'm uncertain about the origin of the "magic number" in the expected . I will reach out to John and inquire about the feasibility of including `n` (sample size) and `event` (number of events) in the expected object of `test-unvalidated-sim_gs_n.R`.

Hi @jdblischak, could you please let us know if it is possible to incorporate `n` and `event` in the expected object of the `test-unvalidated-sim_gs_n.R`?

jdblischak:
> if it is possible to incorporate `n` and `event` in the expected object of the `test-unvalidated-sim_gs_n.R`?

@LittleBeannie Yes, I obtained `expected` by running `constructive::construct(observed)`

> I'm uncertain about the origin of the "magic number" in the expected .

These are backwards compatibility tests. There is nothing magic about them. I cannot refactor the code without having tests to ensure I don't break anything.

LittleBeannie:
> > if it is possible to incorporate `n` and `event` in the expected object of the `test-unvalidated-sim_gs_n.R`?
> 
> @LittleBeannie Yes, I obtained `expected` by running `constructive::construct(observed)`
> 
> > I'm uncertain about the origin of the "magic number" in the expected .
> 
> These are backwards compatibility tests. There is nothing magic about them. I cannot refactor the code without having tests to ensure I don't break anything.

Thank you for your quick response! I encountered an issue while attempting to run `constructive::construct(observed)` directly in the console. It seems that I might have missed the correct way to run the code. Could you please help update the `expected`?

jdblischak (approved):

nanxstats (approved):
Very nice, thanks for the update ✨

PR #214, opened 2024-03-07
Rename `col` to `ctrl` in code examples and tests

nanxstats:
This PR follows #213 to rename `col` to `ctrl` (a more common abbreviation of "control") in code examples and unit tests.

PR #215, opened 2024-03-07
Add `create_cutting_test()` and `multitest()`

jdblischak:
This PR introduces a new function `create_cutting_test()`. Similar to `create_cutting()`, it is a function factory that generates functions that can easily be passed around. This enables us to do various new things.

1. I created a new function `multitest()`. It allows running multiple tests at once on a cut data set. It currently returns a list because the test functions do not return consistent outputs
2. I updated the `sim_gs_n()` argument `test` to accept a list of test functions, one per cutting. This is mostly a proof of concept at the moment. More work will be required before we can support allowing multiple tests per cutting and mixing heterogenous tests (as it is currently written `sim_gs_n()` requires all the test results to be combined into a single output data frame)

I also took the time to clean up some of the roxygen2 docs since I was the one who introduced a lot of the macros.

Lastly, I didn't touch `maxcombo()` in this PR, but you can imagine that we could also use `create_cutting_test()` to be able to pass any arbitrary number of test functions to `maxcombo()`. But just like for `sim_gs_n()` and `multitest()`, this would be facilitated by the test functions having consistent output.


jdblischak:
Rebased onto 9ac40f3 to include #214

nanxstats (approved):
LGTM

PR #221, opened 2024-03-19
Rename function factories create_cut() and create_test()

jdblischak:
Follow-up to #201 and #215.

Based on our internal discussion on how to name these functions, I propose renaming the function factories `create_cut()` and `create_test()`, and renaming their corresponding `sim_gs_n()` arguments to ~~`cuts`~~ `cut` and ~~`tests`~~ `test`.


keaven:
Concept is great. I have not reviewed code or tests.

R/sim_gs_n.R
```diff
@@ -115,8 +117,8 @@
 #'   sample_size = 400,
 #'   enroll_rate = enroll_rate,
 #'   fail_rate = fail_rate,
-#'   test = wlr,
-#'   cutting = list(ia1 = ia1, ia2 = ia2, fa = fa),
+#'   tests = wlr,
```
LittleBeannie, line 120: In both simtrial and gsDesign2, we use singular consistently. If you agree, would it be possible if we change from `tests` to `test` and `cuts` to `cut`?

LittleBeannie (commented):
Thanks, @jdblischak ! I only have 1 minor comment.

R/sim_gs_n.R
jdblischak: I completely agree. Consistency is a good reason to keep the argument names singular. I just pushed the update

R/sim_gs_n.R
```diff
@@ -22,18 +22,20 @@
...
+#'   flexible in the future). Importantly, note that the simulated data set is
+#'   always passed as the first positional argument to each test function
+#'   provided.
+#' @param cut A list of cutting functions created by [create_cut()], see
+#'   examples.
 #' @param seed Random seed.
-#' @param ... Arguments passed to the test function provided by the argument
-#'   `test`.
+#' @param ... Arguments passed to the test function(s) provided by the argument
+#'   `tests`.
```
nanxstats, line 38: `tests` -> `test` for consistency?

nanxstats (changes requested):

R/sim_gs_n.R
jdblischak: Good catch @nanxstats! Fixed in https://github.com/Merck/simtrial/pull/221/commits/d539ae69571753d76e878ec2424ae37424db8c55

nanxstats (approved):
OK let's get this merged! I appreciate the simplification efforts

PR #229, opened 2024-04-18
Enable different tests per cutting for `sim_gs_n()`

jdblischak:
My previous PR #215 enabled passing a different test function to be applied at each cutting orchestrated by `sim_gs_n()`. However, it was quite limited because it only worked when the test functions gave the exact same output (which was rare).

The recent PR #227 from @LittleBeannie standardized the output format of the test functions (see Issue #222 for tracking the progress of this effort). This means we can now combine the results from more diverse test functions.

The trickiest part of this PR was accommodating the tests that return more than one result (`milestone()` returns 2 values for `se` and `maxcombo()` returns 2 values for `z`). To continue returning a single data frame, I had to use list columns.

With this PR, it is now possible to combine the results of `wlr()` (with any arguments), `rmst()`, and `milestone()`. Below is example output where `wlr()` was applied at the first cutting, `rmst()` at the second cutting, and `milestone()` at the third cutting:

```R
head(result)
##      method          parameter  estimation                     se         z analysis cut_date sim_id
## 1       WLR FH(rho=0, gamma=0) -28.1948254               7.521418 -3.748605        1 24.00000      1
## 2      RMST                 20   2.8531309              0.7147367  3.991863        2 32.00000      1
## 3 milestone                 10   0.1752731 0.03495270, 0.03484071 12.078381        3 45.00000      1
## 4       WLR FH(rho=0, gamma=0) -26.8487111               7.721484 -3.477144        1 24.00000      2
## 5      RMST                 20   2.3996924              0.7282783  3.295021        2 32.00000      2
## 6 milestone                 10   0.1173894 0.03543263, 0.03491216  5.457930        3 46.21933      2
##     n event
## 1 400   229
## 2 400   295
## 3 400   355
## 4 400   241
## 5 400   290
## 6 400   350

str(result)
## 'data.frame':	9 obs. of  10 variables:
##  $ method    : chr  "WLR" "RMST" "milestone" "WLR" ...
##  $ parameter : chr  "FH(rho=0, gamma=0)" "20" "10" "FH(rho=0, gamma=0)" ...
##  $ estimation: num  -28.195 2.853 0.175 -26.849 2.4 ...
##  $ se        :List of 9
##   ..$ : num 7.52
##   ..$ : num 0.715
##   ..$ : num  0.035 0.0348
##   ..$ : num 7.72
##   ..$ : num 0.728
##   ..$ : num  0.0354 0.0349
##   ..$ : num 7.5
##   ..$ : num 0.742
[...]

Issue #205, opened 2024-02-23
Implement parallel adaptor in `sim_gs_n()`


nanxstats:
We would like `sim_gs_n()` to have a parallel adaptor implemented.

So it achieves feature parity as `sim_fixed_n()` which uses `%dofuture%` (xref: https://github.com/Merck/simtrial/pull/110).

Maybe **after** the related APIs to be called in `sim_gs_n()` are in a more finalized state.

nanxstats:
Fixed by #249 and #252

PR #230, opened 2024-04-19
Enhancement for doFuture implementation and stable parallel RNG in sim_gs_n

cmansch:
Implementation for `sim_gs_n()` to have a parallel adaptor similar to `sim_fixed_n()` (xref: https://github.com/Merck/simtrial/pull/110).
A seed is now set prior to the function call instead of passed as an argument.
Function now uses `%douture%` to run iterations of `n_sim` in parallel. 

Addresses open issue (xref: https://github.com/Merck/simtrial/issues/205).

Sequential backend still available (and default) to avoid any delays in development.

Please update "Remove seed argument from `sim_gs_n()` in (xref: https://github.com/Merck/gsDesign2/issues/288) as part of the implementation. 

PR will be updated again following approval of PR (xref: https://github.com/Merck/simtrial/pull/229) as it will need to combine updates to `sim_gs_n` from that PR.

### DIFF
--- a/man/sim_gs_n.Rd
+++ b/man/sim_gs_n.Rd
@@ -224,13 +224,19 @@ sim_gs_n(
 # Example 8: At IA1, we conduct 3 tests, LR, WLR with fh(0, 0.5), and RMST test.
 # At IA2, we conduct 2 tests, LR and WLR with early zero (6).
 # At FA, we conduct 2 tests, LR and milestone test.
-ia1_test <- list(test1 = create_test(wlr, weight = fh(rho = 0, gamma = 0)),
-                 test2 = create_test(wlr, weight = fh(rho = 0, gamma = 0.5)),
-                 test3 = create_test(rmst, tau = 20))
-ia2_test <- list(test1 = create_test(wlr, weight = fh(rho = 0, gamma = 0)),
-                 test2 = create_test(wlr, weight = early_zero(6)))
-fa_test <- list(test1 = create_test(wlr, weight = fh(rho = 0, gamma = 0)),
-                 test3 = create_test(milestone, ms_time = 20))
+ia1_test <- list(
+  test1 = create_test(wlr, weight = fh(rho = 0, gamma = 0)),
+  test2 = create_test(wlr, weight = fh(rho = 0, gamma = 0.5)),
+  test3 = create_test(rmst, tau = 20)
+)
+ia2_test <- list(
+  test1 = create_test(wlr, weight = fh(rho = 0, gamma = 0)),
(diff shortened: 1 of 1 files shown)


R/sim_gs_n.R
```diff
@@ -112,16 +112,15 @@
...
-#' )
+# sim_gs_n(
+#   n_sim = 3,
+#   sample_size = 400,
+#   enroll_rate = enroll_rate,
+#   fail_rate = fail_rate,
+#   test = wlr,
+#   cut = list(ia1 = ia1, ia2 = ia2, fa = fa),
+#   weight = fh(rho = 0, gamma = 0)
+# )
```
jdblischak, line 123: Please add back the apostrophe (`#'`) as this removed this example from the documentation. Also note that if you are using RStudio, you can execute this code with Ctrl-Enter without having to uncomment it.

tests/testthat/test-unvalidated-sim_gs_n.R
```diff
@@ -4,14 +4,14 @@
 # See helper-sim_gs_n.R for helper functions
 
 test_that("Test 1: regular logrank test", {
+  # set.seed(2024)
```
jdblischak, line 7: This was unfortunate timing as all the `sim_gs_n()` tests were commented out. Once #229 is merged, you can rebase on top of that, and then the tests will run.

tests/testthat/test-unvalidated-sim_gs_n.R
```diff
@@ -272,19 +272,19 @@ test_that("sim_gs_n() accepts different tests per cutting", {
...
+  #     n_sim = 3,
+  #     sample_size = 400,
+  #     enroll_rate = test_enroll_rate(),
+  #     fail_rate = test_fail_rate(),
+  #     test = list(wlr_cut1, wlr_cut2),
+  #     cut = test_cutting()
+  #   ),
+  #   "If you want to run different tests at each cutting"
+  # )
+  expect_equal(1 + 1, 2)
```
jdblischak, line 289: In the future, instead of commenting out a test, please add `skip("The reason the test needs to be skipped")`. This keeps the `git diff`'s nicer. In other words, it is easier to see what needed to be change to fix a test if it wasn't completely commented out previously.

jdblischak (changes requested):
Thanks @cmansch! Looking forward to this!

jdblischak:
My PR #229 has been merged. @cmansch could you please update your PR? I recommend fixing the roxygen2 comments first since that should reduce the number of merge conflicts you'll have to address manually

PR #237, opened 2024-04-24
236 milestone take the un squared version of the current z statistics

### LINKED ISSUE #236: `milestone()`: take the un-squared version of the current Z statistics
The equation (3) of Klein, et al (2007) published at Stat in Medicine has a squared version of the test statistics. We will take the un-squared version to provide a directional Z test.
image

LittleBeannie:
Closes #236

### DIFF
--- a/R/milestone.R
+++ b/R/milestone.R
@@ -23,13 +23,17 @@
 #'   - `event` - Event indicator.
 #'   - `treatment` - Grouping variable.
 #' @param ms_time Milestone analysis time.
-#'
+#' @param test_type Method to build the test statistics.
+#' There are 2 options:
+#'   - `"native"`: a native approach by dividing the KM survival difference by its standard derivations.
+#'   - `"log-log"`: a log-log transformation of the survival, see equation (3) of
+#'   Klein, J. P., Logan, B., Harhoff, M., & Andersen, P. K. (2007).
 #' @return A list frame containing:
-#' - `method` - The method, always `"milestone"`.
-#' - `parameter` - Milestone time point.
-#' - `estimation` - Survival difference between the experimental and control arm.
-#' - `se` - Standard error of the control and experimental arm.
-#' - `z` - Test statistics.
+#'   - `method` - The method, always `"milestone"`.
+#'   - `parameter` - Milestone time point.
+#'   - `estimation` - Survival difference between the experimental and control arm.
+#'   - `se` - Standard error of the control and experimental arm.
+#'   - `z` - Test statistics.
 #'
 #' @references
 #' Klein, J. P., Logan, B., Harhoff, M., & Andersen, P. K. (2007).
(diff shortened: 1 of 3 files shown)


nanxstats (approved):
I see a unit test is skipped for now - would be great to unskip it later after the hypothesis testing syntax is updated.

jdblischak:
> I see the "multitest" unit test is skipped for now - would be great to unskip it later after the hypothesis testing syntax is updated.

I don't understand why this test was skipped. It doesn't run the updated `milestone()` function, so it should have been unaffected by this PR. I just ran it locally, and it passed

nanxstats:
> I don't understand why this test was skipped. It doesn't run the updated `milestone()` function, so it should have been unaffected by this PR. I just ran it locally, and it passed

🤔 That would be a good question for @LittleBeannie

R/milestone.R
```diff
@@ -23,13 +23,18 @@
 #'   - `event` - Event indicator.
 #'   - `treatment` - Grouping variable.
 #' @param ms_time Milestone analysis time.
-#'
+#' @param test_type Method to build the test statistics.
+#' There are 2 options:
+#'   - `"native"`: a native approach by dividing the KM survival difference by its standard derivations,
```
jdblischak, line 28: Is the approach "native" or "naive"? The string passed to `test_type` below is `"naive"`

R/milestone.R
```diff
@@ -23,13 +23,18 @@
...
 #' @return A list frame containing:
-#' - `method` - The method, always `"milestone"`.
-#' - `parameter` - Milestone time point.
-#' - `estimation` - Survival difference between the experimental and control arm.
-#' - `se` - Standard error of the control and experimental arm.
-#' - `z` - Test statistics.
+#'   - `method` - The method, always `"milestone"`.
+#'   - `parameter` - Milestone time point.
+#'   - `estimation` - Survival difference between the experimental and control arm.
+#'   - `se` - Standard error of the control and experimental arm.
```
jdblischak, line 36: The `se` result was changed by this PR, but the documentation did not change. Previously it returned two standard error values (from `fit_res$std.err`), but now it only returns a single standard error. Is is the standard error of the control arm, experimental arm, or some combination? We should update the documentation for this return value

jdblischak (commented):
Overall this PR had a much bigger impact than I had anticipated. I assumed it was only changing the Z statistics in `z`. However, it also changed `se` to only return a single value instead of two. And the default `estimate` is also changed by the new `test_type` of `"log-log"`. I was able to fix `estimate` for my `sim_gs_n()` tests by adding `test_type = "naive"`.

We should add some developer tests for `milestone()` so that it is more transparent how a PR is effecting the output.

R/milestone.R
nanxstats: Probably a typo - should be "naive" cc @LittleBeannie

R/milestone.R
nanxstats: Agreed, this should be a quick fix for @LittleBeannie

PR #242, opened 2024-05-02
Require a single formula syntax for rmst()

### LINKED ISSUE #220: Formula parsing: match by argument names instead of by position
This is a follow up on https://github.com/Merck/simtrial/pull/216 about the new formula interface for `rmst()`.

I agree that we should not evaluate the `Surv()` call and only extract the variable names from the formula. I do think that having a single, canonical syntax would be beneficial, such as `Surv(month, status) ~ trt)`, and safeguard common pitfalls with essential enforcement so it's difficult for users to write the wrong code without realizing.

Aligned with this spirit, @jdblischak's [comment](https://github.com/Merck/simtrial/pull/216#issuecomment-2000076501) is right on point that matching by position when parsing the formula is dangerous as people could write legal `Surv()` calls with shuffled argument positions such as `Surv(event = status, time = month) ~ trt` without realizing the matched variables will be wrong.

So, here is an _ad hoc_ "solution" to the problem. I wonder if either rlang or [adv-r first edition](http://adv-r.had.co.nz/Expressions.html) has better, direct alternatives.

``` r
# Copied from <https://github.com/cran/stackoverflow/blob/master/R/match.call.default.R>
match_call_defaults <- function(
    definition = sys.function(sys.parent()),
    call = sys.call(sys.parent()),
    expand.dots = TRUE,
    envir = parent.frame(2L)) {
  call <- match.call(definition, call, expand.dots, envir)
  formals <- formals(definition)

  if (expand.dots && "..." %in% names(formals)) formals[["..."]] <- NULL
  for (i in setdiff(names(formals), names(call))) call[i] <- list(formals[[i]])

  match.call(definition, call, TRUE, envir)
}

rmst <- function(data, formula = NULL) {
  formula <- as.formula(formula)
  var_label_group <- attr(stats::terms(formula), "term.labels")[1]
  surv_call <- attr(stats::terms(formula), "variables")[[2]]
  surv_call_match <- match_call_defaults(definition = function(time, event) {}, surv_call)
  var_label_tte <- as.character(surv_call_match$time)
  var_label_event <- as.character(surv_call_match$event)

  c(
    "var_label_tte" = var_label_tte,
    "var_label_event" = var_label_event,
    "var_label_group" = var_label_group
  )
}

# rhs - will always and only match the very first term
rmst(formula = Surv(month, status) ~ trt + trt2)
#>   var_label_tte var_label_event var_label_group
#>         "month"        "status"           "trt"

# lhs - canonical case works
rmst(formula = Surv(month, status) ~ trt)
#>   var_label_tte var_label_event var_label_group
#>         "month"        "status"           "trt"

# lhs - all named arguments work
rmst(formula = Surv(event = status, time = month) ~ trt)
#>   var_label_tte var_label_event var_label_group
#>         "month"        "status"           "trt"

# lhs - abbreviate first argument name works
rmst(formula = Surv(month, event = status) ~ trt)
#>   var_label_tte var_label_event var_label_group
#>         "month"        "status"           "trt"

# lhs - partial positional matching - not common but works
rmst(formula = Surv(event = status, month) ~ trt)
#>   var_label_tte var_label_event var_label_group
#>         "month"        "status"           "trt"

# lhs - giving more or fewer arguments will not work - can use some handling
rmst(formula = Surv(month) ~ trt)
#>   var_label_tte var_label_group
#>         "month"           "trt"
rmst(formula = Surv(month, month2, status) ~ trt)
#> Error in match.call(definition, call, expand.dots, envir): unused argument (status)
```

jdblischak:
Closes #220 

Follow-up to #216 

The `formula` argument of `rmst()` now requires a single syntax, `Surv(tte, event) ~ group)`. However, to be consistent with user expectations, it is aware of the argument names of `Surv()`, so it is flexible enough to handle valid input such as `Surv(event = event, tte) ~ group`. Thanks to @nanxstats for the argument parsing code.

Below are examples of valid and invalid formulas:

```R
# valid
Surv(tte, event) ~ group)
Surv(event = event, time = tte) ~ group)
Surv(tte, event = event) ~ group)
Surv(event = event, tte) ~ group)

# invalid
Surv(tte | event) ~ group)
~ Surv(tte, evntd, group)
tte ~ event + group
~ tte + event + group
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: simtrial
 Type: Package
 Title: Clinical Trial Simulation
-Version: 0.4.1
+Version: 0.4.1.1
 Authors@R: c(
     person("Keaven", "Anderson", email = "keaven_anderson@merck.com", role = c("aut")),
     person("Yujie", "Zhao", email = "yujie.zhao@merck.com", role =  c("ctb","cre")),
--- a/R/rmst.R
+++ b/R/rmst.R
@@ -26,23 +26,22 @@
 #' @param var_label_event Column name of the event variable.
 #' @param var_label_group Column name of the grouping variable.
 #' @param formula (default: `NULL`) A formula that indicates the TTE, event, and
-#'   group variables (in that exact order; see Details below). This is an
-#'   alternative to specifying the variables as strings. If a formula is
-#'   provided, the values passed to `var_label_tte`, `var_label_event`, and
-#'   `var_label_group` are ignored.
+#'   group variables using the syntax `Surv(tte, event) ~ group)` (see Details
+#'   below for more information). This is an alternative to specifying the
+#'   variables as strings. If a formula is provided, the values passed to
+#'   `var_label_tte`, `var_label_event`, and `var_label_group` are ignored.
 #' @param reference A group label indicating the reference group.
 #' @param alpha Type I error.
 #'
 #' @details
-#' The argument `formula` is provided as a convenience to easily specify the TTE,
-#' event, and grouping variables. Note however that only the order of the three
-#' variables is actually used by the underlying function. Any functions applied
-#' in the formula are ignored, and thus should only be used for documenting your
-#' intent. For example, you can use the syntax from the survival package
-#' `Surv(tte | event) ~ group` to highlight the relation between the TTE and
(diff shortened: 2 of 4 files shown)


nanxstats:
You might want to merge this shortly after we release 0.4.1.

PR #247, opened 2024-05-06
Replace rlang::is_installed() with base::requireNamespace()

### LINKED ISSUE #246: Replace `rlang::is_installed()` with a lightweight solution
I used `rlang::is_installed()` and introduced rlang as a hard dependency in #245 to safeguard code examples, tests, and vignettes and get the package pass the `noSuggests` check.

It would be ideal if we can use a lightweight solution here and remove the dependency.

@jdblischak suggested `requireNamespace()`. @yihui also mentioned `xfun::loadable()`.

jdblischak:
Closes #246 

This PR removes the direct dependency on {rlang} by replacing `rlang::is_installed()` with `base::requireNamespace(quietly = TRUE)`. I was originally going to create an internal function `is_installed()`, but even that seems uncessary since this code is always hidden from the user in the examples and vignettes.

I tested locally that `requireNamespace()` behaves as required:

```R
remove.packages("dplyr")
library("simtrial")
requireNamespace("dplyr", quietly = TRUE)
.Last.value
## [1] FALSE
example("get_cut_date_by_event")
devtools::check(force_suggests = FALSE) # Sets _R_CHECK_FORCE_SUGGESTS_⁠ to FALSE
## N  checking package dependencies (11s)
##    Package suggested but not available for checking: 'dplyr'
```

xref: https://github.com/Merck/simtrial/pull/245, https://github.com/Merck/gsDesign2/pull/307, https://github.com/Merck/simtrial/pull/195#pullrequestreview-1876654204, https://github.com/Merck/simtrial/pull/201


nanxstats (approved):
Great! Thanks for restoring the dependency state to where it should be.

PR #248, opened 2024-05-07
Run roxygen2

nanxstats:
This PR runs roxygen2 to refresh the `.Rd` files, follow the updates in #229.


LittleBeannie (approved):
Thanks, @nanxstats !

PR #252, opened 2024-05-28
Fixing sim_fixed_n() parallel results return

cmansch:
This update addresses two open issues: https://github.com/Merck/simtrial/issues/250 and https://github.com/Merck/simtrial/issues/251. 

In order to return the results from each `%dofuture%` loop, the last object must return some value. The `setDF()` call previously used resulted in the results not being passed back from the loop when run with a parallel backend. This error was hidden in the results as the loop used the `.errorhandling = "pass"` instead of `"stop"`. 

The parallel vignette was also updated to indicate that `sim_gs_n()` supports the same backend implementation as `sim_fixed_n()`.

Issue #251, opened 2024-05-28
Example 4 of `sim_fixed_n` does not work


LittleBeannie:
Hi @cmansch , it looks the last example of `sim_fixed_n` does not work, which is related to parallel computation. Could you please help take a look at it? 

Also looping in @jdblischak. It looks like the stylize codes update may change how the results returned. https://github.com/Merck/simtrial/blob/04b36d068e4a3b9b6a8a6c739fa9576be2a556fc/R/sim_fixed_n.R

![image](https://github.com/Merck/simtrial/assets/43153957/debfbe65-be69-4510-badc-4778ae168bb0)

jdblischak:
Interesting. I think we stumbled across the same problem a few weeks in @cmansch's PR to parallelize `sim_gs_n()` (https://github.com/Merck/simtrial/pull/249). When he updated `sim_fixed_n()` to use `.errorhandling = "stop"`, the vignettes started failing with the error `object 'results' not found` (https://github.com/Merck/simtrial/pull/249#issuecomment-2115963072), just like above. Thus fixing this Issue should also fix the issue to update the error handling of `sim_fixed_n()` (https://github.com/Merck/simtrial/issues/250).

I'm going to investigate more now. But the current situation is that the parallel executions of `sim_fixed_n()` in the examples and vignettes only succeed because the setting `.errorhandling = "pass"` does not fail when errors occur. So now the question is: did parallel execution of `sim_fixed_n()` work when it was first introduced in https://github.com/Merck/simtrial/pull/110? And if yes, what changed?

jdblischak:
I'm not sure parallel execution of `sim_fixed_n()` ever worked. At least not with the latest versions of {future}, {doParallel}, etc. I checked out the merge commit of PR #110, 341f77f0a598dc6d638bd5c48746952a7db88255, and it fails with the same error.

```R
git checkout 341f77f0a598dc6d638bd5c48746952a7db88255
git log -n 1 --oneline
## 341f77f (HEAD) Merge pull request #110 from cmansch/main
Rscript -e 'devtools::load_all(); library("future"); plan("multisession", workers=2); sim_fixed_n(n_sim = 10)'
## ℹ Loading simtrial
## Using 2 cores with backend multisession
##           message                                   call
## result.1  "unused argument (rho_gamma = rho_gamma)" expression
## result.2  "unused argument (rho_gamma = rho_gamma)" expression
## result.3  "unused argument (rho_gamma = rho_gamma)" expression
[...]

PR #249, opened 2024-05-13
Adding %doFuture% parallel framework for the sim_gs_n.R code and upda…

cmansch:
Implementation for sim_gs_n() to have a parallel adaptor similar to sim_fixed_n() (xref: https://github.com/Merck/simtrial/pull/110).
A seed is now set prior to the function call instead of passed as an argument.
Function now uses %doFuture% to run iterations of n_sim in parallel.

Addresses open issue (xref: https://github.com/Merck/simtrial/issues/205).

Sequential backend still available (and default) to avoid any delays in development.

Please update "Remove seed argument from sim_gs_n() in (xref: https://github.com/Merck/gsDesign2/issues/288) as part of the implementation.

New pull request created after (xref: https://github.com/Merck/simtrial/pull/229) merged.
See old PR as (xref: https://github.com/Merck/simtrial/pull/230)

PR #253, opened 2024-05-29
Import `nbrOfWorkers()` from {future} into `NAMESPACE`

jdblischak:
While evaluating #251 and #252, I discovered that the function `nbrOfWorkers()` is missing from the package `NAMESPACE`. Below is a minimal reprex to demonstrate the problem:

```R
library("simtrial")
packageVersion("simtrial")
## [1] ‘0.4.1.4’
future::plan("multisession", workers = 2)
sim_fixed_n(n_sim = 10)
## Error in nbrOfWorkers() : could not find function "nbrOfWorkers"
```

This `NAMESPACE` problem was missed by `R CMD check` because I wrongly added `"nbrOfWorkers"` to `utils::globalVariables()` in #114, assuming it was a variable defined in this package.

https://github.com/Merck/simtrial/blob/75297a12f0162d8fa6388ae23f627fd0ba12804b/R/global.R#L46

After removing it from `utils::globalVariables()`, `R CMD check` properly identifies the problem:

```
N  checking R code for possible problems (6.5s)
   sim_fixed_n: no visible global function definition for 'nbrOfWorkers'
   sim_gs_n: no visible global function definition for 'nbrOfWorkers'
     nbrOfWorkers
```

And after importing `nbrOfWorkers()` into the package namespace, the example above now works:

```R
# after
library("simtrial")
packageVersion("simtrial")
## [1] ‘0.4.1.5’
future::plan("multisession", workers = 2)
sim_fixed_n(n_sim = 10)
# works
```

xref: #110, #249


nanxstats (approved):

cmansch (approved):
Good catch.

Issue #260, opened 2024-07-17
A potential bug of parallel computation in `sim_gs_n`


LittleBeannie:
The following lines of codes gives me an error message when I run it parallelly. However, the sequential processing is error-free.
```
ratio <- 1

enroll_rate <- gsDesign2::define_enroll_rate(duration = c(2, 2, 8),
                                             rate = c(1, 2, 3))

fail_rate <- gsDesign2::define_fail_rate(duration = c(4, Inf), 
                                         fail_rate = log(2) / 12,
                                         hr = c(1, .6),
                                         dropout_rate = .001)

alpha <- 0.025
beta <- 0.1

upper <- gsDesign2::gs_spending_bound
upar <- list(sf = gsDesign::sfLDOF, total_spend = alpha)
test_upper <- rep(TRUE, 2)

lower <- gsDesign2::gs_spending_bound
lpar <- list(sf = gsDesign::sfLDOF, total_spend = beta)
test_lower <- c(TRUE, FALSE)
binding <- FALSE

info_frac = NULL
analysis_time = c(24, 36)

x <- gsDesign2::gs_design_ahr(enroll_rate = enroll_rate, fail_rate = fail_rate, 
                              alpha = alpha, beta = beta, ratio = ratio,
                              info_frac = info_frac, analysis_time = analysis_time, 
                              upper = upper, upar = upar, test_upper = test_upper,
                              lower = lower, lpar = lpar, test_lower = test_lower,
                              binding = binding) |> gsDesign2::to_integer()

ia_cut <- simtrial::create_cut(planned_calendar_time = x$analysis$time[1])
fa_cut <- simtrial::create_cut(planned_calendar_time = x$analysis$time[2])

future::plan("multisession", workers = 8)
simtrial::sim_gs_n(
  n_sim = 1e2,
  sample_size = x$analysis$n[2],
  enroll_rate = x$enroll_rate,
  fail_rate = x$fail_rate,
  test = simtrial::wlr,
  cut = list(ia = ia_cut, fa = fa_cut),
  weight = simtrial::fh(rho = 0, gamma = 0))
future::plan("sequential")
```

Following is the error message.
```
Using 8 cores with backend multisession
Error in eval(expr, p) : object 'x' not found
In addition: There were 50 or more warnings (use warnings() to see the first 50)
> warnings()
Warning messages:
1: In input_check_scalar(planned_calendar_time) :
  restarting interrupted promise evaluation
```

[...]

PR #261, opened 2024-07-17
Fixing Bug for Cut Functions in Parallel

cmansch:
Cut functions that use globals were not resolving due to a lazy evaluation within a foreach/doFuture framework. The solution is to force the `create_cut` function to evaluate its arguments. 
Addresses #260. 
Pull request has a single line of code updated from previous versions to fix.

### DIFF
--- a/R/sim_gs_n.R
+++ b/R/sim_gs_n.R
@@ -271,7 +271,6 @@ sim_gs_n <- function(
   ans <- foreach::foreach(
     sim_id = seq_len(n_sim),
     test = replicate(n=n_sim, expr=test, simplify = FALSE),
-    cut = replicate(n=n_sim, expr=cut, simplify = FALSE),
     .combine = "rbind",
     .errorhandling = "stop",
     .options.future = list(seed = TRUE)
@@ -361,6 +360,7 @@ sim_gs_n <- function(
 #' # Cut the trial data
 #' cutting(trial_data)
 create_cut <- function(...) {
+  lapply(X = list(...), FUN = fo)
   function(data) {
     get_analysis_date(data, ...)
   }
--- a/R/sim_gs_n.R
+++ b/R/sim_gs_n.R
@@ -360,7 +360,7 @@ sim_gs_n <- function(
 #' # Cut the trial data
 #' cutting(trial_data)
 create_cut <- function(...) {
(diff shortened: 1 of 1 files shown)


jdblischak (commented):
I installed this locally with `remotes::install_github("cmansch/simtrial-test")`, but I'm still getting the same error on Windows.

cmansch:
> I installed this locally with `remotes::install_github("cmansch/simtrial-test")`, but I'm still getting the same error on Windows.

I will look for a more robust solution.

R/sim_gs_n.R
```diff
@@ -360,6 +360,7 @@ sim_gs_n <- function(
 #' # Cut the trial data
 #' cutting(trial_data)
 create_cut <- function(...) {
+  lapply(X = list(...), FUN = fo)
```
jdblischak, line 363:
```suggestion
  lapply(X = list(...), FUN = force)
```

jdblischak (approved):

LittleBeannie:
Thank you, @cmansch and @jdblischak for fix it timely!

R/sim_gs_n.R
```diff
@@ -360,6 +360,7 @@ sim_gs_n <- function(
 #' # Cut the trial data
 #' cutting(trial_data)
 create_cut <- function(...) {
+  lapply(X = list(...), FUN = force)
```
yihui, line 363: Just a quick technical note: I'm not sure if everyone knows that `force()` is simply semantic sugar (as its help page `?force` says). All the sugar  helps with is perhaps code readability. Technically, if you want to avoid the lazy/delayed evaluation of function arguments, you can simply use the arguments earlier by placing their names where you want them to be evaluated, e.g., here you can simply write `list(...)` instead of `lapply(X = list(...), FUN = force)`.

jdblischak: Personally I like the increased readability of `force()`. As it is, this isn't a widely known feature. Also I added a comment in #262 to explain the purpose. I'd be open to switching to the minimal `list(...)` once the comment is available for context

yihui: Either way is fine to me. I don't have a strong opinion. Personally I slightly prefer low calories :)

PR #263, opened 2024-07-23
Use KaTeX for pkgdown math rendering

nanxstats:
## What's in this PR

This PR brings back beautiful math rendering to the pkgdown vignettes by switching the math rendering engine to KaTeX supported in pkgdown 2.1.0.

## Before and after

This is what an [example vignette section](https://merck.github.io/simtrial/articles/rmst.html#estimation-of-rmst-in-a-single-arm-at-a-single-time-point) looks like before, current, and after this patch.

### pkgdown 2.0.9 (old default, MathJax)

<details>
<summary>
Click to expand screenshot
</summary>

mathjax

</details>

### pkgdown 2.1.0 (new default, MathML)

<details>
<summary>
Click to expand screenshot
</summary>

mathml

</details>

### pkgdown 2.1.0 + KaTeX

<details>
<summary>
Click to expand screenshot
</summary>

katex

</details>

## Method

To make this work, I shipped the KaTeX math fonts using the [pkgdown customization](https://pkgdown.r-lib.org/articles/customise.html) mechanism `pkgdown/extra.css` and `pkgdown/assets/fonts/`. This is because pkgdown 2.1.0 did not bundle them (not sure why) and the default math font rendering is subpar.

## Alternative paths

Apparently, MathML doesn't render very well for our formulas. Although we can choose the old default MathJaX, the font bundling situation is a bit more challenging to fix using the same customization approach as the fonts were defining in JS (I successfully converted them to CSS). Considering that KaTeX is faster and simpler, I think it's the optimal choice.

xref: https://github.com/keaven/gsDesign/pull/161, https://github.com/Merck/gsDesign2/pull/437


jdblischak (approved):

LittleBeannie (approved):
Thanks, @nanxstats !

PR #267, opened 2024-07-29
266 Correct the output column name of `counting_process()`

LittleBeannie:


R/counting_process.R
```diff
@@ -97,50 +97,53 @@ counting_process <- function(x, arm) {
...
   # Variance of o_minus_e
-  ans[, var_o_minus_e := (n_risk_tol - n_risk_trt) *
-    n_risk_trt * events * (n_risk_tol - events) /
-    n_risk_tol^2 / (n_risk_tol - 1)]
+  ans[, var_o_minus_e := (n_risk_total - n_risk_trt) *
+    n_risk_trt * event_total * (n_risk_total - event_total) /
+    n_risk_total^2 / (n_risk_total - 1)]
 
   setDF(ans)
+  class(ans) <- c("counting_process", class(ans))
```
jdblischak, line 146: What is the purpose of adding the class `"counting_process"` to the returned object? There are no S3 methods defined for this class. Until S3 methods are defined, I think we should keep things simple and not create unused classes.

jdblischak (changes requested):

R/counting_process.R
LittleBeannie: Thanks for sharing your insights, @jdblischak! It is a request from Keaven. We will develop some S3 method for `counting_process` later in the future.

Here are some further information.
- If we have a counting_process object, then we can directly get the test statistics.
- If we have original TTE data, we should transfer it to a counting_process object and then to test.

jdblischak: > We will develop some S3 method for `counting_process` later in the future.

My argument is that adding the class to the output object should be added in the same future PR that adds the methods. This PR only renames the output columns, and would be cleaner if it was limited to only that change.

jdblischak: Summary from Teams discussion: `counting_process` class is needed now even though no S3 methods yet exist

jdblischak (approved):

PR #268, opened 2024-07-29
265 Summarize `sim_gs_n`

LittleBeannie:


R/sim_gs_n.R
```diff
@@ -329,6 +329,13 @@ sim_gs_n <- function(
 
     ans_1sim
   }
+
+  test_method <- ans$method[ans$sim_id == 1]
+  if (all(substr(test_method, 1, 3) == "WLR")) {
+    class(ans) <- c("simtrial_gs_wlr", class(ans))
+    attr(ans, "method") <- unique(ans$parameter[ans$sim_id == 1])
+  }
```
jdblischak, line 337: It's my understanding that this class of `"simtrial_gs_wlr"` will be applied only when the test at each cutting is always `"WLR"`. Before we start adding classes and methods to {simtrial}, I think it would be good to discuss the plan for all the various classes and methods we envision adding. That way we can think through all the possibilities and then name them accordingly

jdblischak (changes requested):

R/sim_gs_n.R
LittleBeannie: It's my understanding that this class of "simtrial_gs_wlr" will be applied only when the test at each cutting is always "WLR". [YZ: yes, you are correct.]

I think it would be good to discuss the plan for all the various classes and methods we envision adding. [YZ: to my point of view, there are 2 major tests, one is WLR and the other is MaxCombo. If we need further classification of WLR, then we could classify by weighting function, such FH, MB, ect..]

That way we can think through all the possibilities and then name them accordingly [Sure, let's discuss further next week.]

R/global.R
```diff
@@ -23,14 +23,20 @@
...
+    "bound",
     "ex1_delayed_effect",
-    "N",
     "cte",
+    "cross_lower",
+    "cross_upper",
     "dropout_rate",
     "dropout_time",
     "duration",
+    "ends_with",
```
nanxstats, line 39: `ends_with()` is a function from tidyselect, not a non-standard evaluation variable name. So `#' @importFrom tidyselect ends_with` in the roxygen documentation is the right solution.

LittleBeannie: Thanks for pointing it out, @nanxstats ! I deleted the `ends_with` in the latest commit.

nanxstats: Perhaps I didn't point it out clearly enough - you need to import it to avoid check notes.

keaven (approved):

DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: simtrial
 Type: Package
 Title: Clinical Trial Simulation
-Version: 0.4.1.6
+Version: 0.4.1.7
```
jdblischak, line 4: The current version is 0.4.1.8

https://github.com/Merck/simtrial/blob/69960550e90e11ff88bb210586eb48523403f32e/DESCRIPTION#L4

R/as_gt.R
```diff
@@ -0,0 +1,144 @@
...
+#'
+#' @param x A summary object of a fixed or group sequential design.
+#' @param ... Additional arguments (not used).
+#'
+#' @return A gt table.
+#'
+#' @export
+as_gt <- function(x, ...) {
+  UseMethod("as_gt", x)
+}
```
jdblischak, line 29: I confirmed that we also export `as_gt()` in gsDesign2

https://github.com/Merck/gsDesign2/blob/a77b22f114eba302f199e11941e649a4f6aee739/NAMESPACE#L13

R/as_gt.R
```diff
@@ -0,0 +1,144 @@
...
+#'   enroll_rate = enroll_rate, fail_rate = fail_rate,
+#'   ratio = ratio, alpha = alpha, beta = beta,
+#'   analysis_time = c(12, 24, 36),
+#'   upper = gsDesign2::gs_spending_bound,
+#'   upar = list(sf = gsDesign::sfLDOF, total_spend = alpha),
+#'   lower = gsDesign2::gs_b,
+#'   lpar = rep(-Inf, 3))
+#'
+#' # Define cuttings of 2 IAs and 1 FA
+#' ia1_cut <- simtrial::create_cut(target_event_overall = ceiling(design$analysis$event[1]))
```
jdblischak, line 78: We shouldn't need to prefix with `simtrial::` when these examples are part of the {simtrial} package

R/sim_gs_n.R
jdblischak: I don't believe we ever had time to discuss this at a group meeting. Let's proceed, and we can always adjust the class names later

R/summary.R
```diff
@@ -0,0 +1,220 @@
...
+          ) |>
+          dplyr::mutate(cross_upper = -z >= upper_bound)
+      )
+    } else {
+      suppressMessages(
+        sim_tbl <- object |>
+          dplyr::left_join(
+            design$bound |>
+              dplyr::select(analysis, z, bound) |>
+              tidyr::pivot_wider(values_from = z, names_from = bound) |>
```
jdblischak, line 141: {dplyr} and {tidyr} are only in Suggests. Is the plan for me to convert this function to use {data.table}?

jdblischak (commented):
I think this PR is a great candidate to use the "Squash and Merge" feature with a single informative commit like "Add as_gt() and summary() methods for simtrial_gs_wlr"

DESCRIPTION
LittleBeannie: Updated

R/as_gt.R
LittleBeannie: Exactly. If we have `as_gt` in `simtrial`, we should have `... |> simtrial::as_gt()` to distinguish between `gsDesign2::as_gt()`.

Do you happen to have an alternative way? such as re-naming `as_gt`?

R/as_gt.R
LittleBeannie: Updated

R/summary.R
LittleBeannie: Yes, please.

R/as_gt.R
jdblischak: No, I can't think of a better way. While you could rename `as_gt()`, I think it is more consistent to keep the same name.

jdblischak (approved):

R/as_gt.R
LittleBeannie: Agree!

Issue #259, opened 2024-07-12
Add the info to the output of `sim_gs_n()`


LittleBeannie:
The PR will fix the 2 issues below.

1. The weighting function (such as `fh_weight`, `mb_weight` and `early_zero_weight`) now all gives a column of `"weight"`.
2. The info and info0 are added as output of `sim_gs_n`.

LittleBeannie:
To test the calculation of the simulated info and info0, I compared the numbers with the asymptotic results.

**Comparision 1: Regular logrank test**
![image](https://github.com/user-attachments/assets/a139a421-4280-4546-8ccd-48375592faaf)
```
# asymptotic info
x <- gs_design_wlr(enroll_rate = enroll_rate,
                   fail_rate = fail_rate,
                   ratio = 1,
                   alpha = 0.025, beta = 0.2,
                   weight = function(x, arm0, arm1) {wlr_weight_fh(x, arm0, arm1, rho = 0, gamma = 0)},
                   upper = gs_spending_bound,
                   upar = list(sf = gsDesign::sfLDOF, total_spend = 0.025),
                   lower = gs_b,
                   lpar = rep(-Inf, 1),
                   analysis_time = 36) |> to_integer()

y <- gs_design_ahr(enroll_rate = enroll_rate,
                   fail_rate = fail_rate,
                   ratio = 1,
                   alpha = 0.025, beta = 0.2,
                   upper = gs_spending_bound,
                   upar = list(sf = gsDesign::sfLDOF, total_spend = 0.025),
                   lower = gsDesign2::gs_b,
                   lpar = rep(-Inf, 1),
                   analysis_time = 36) |> to_integer()

# simulated info
registerDoFuture()
registerDoRNG()
plan("multisession", workers = 24)
set.seed(2024)
sim_res <- foreach(
  sim_id = seq_len(1e4),
  .combine = "rbind",
  .errorhandling = "pass"
) %dorng% {
  do.call(cbind,
          sim_pw_surv(n = round(x$analysis$n, 0),
                      stratum = data.frame(stratum = "All", p = 1),
                      block = c(rep("control", 2), rep("experimental", 2)),
                      enroll_rate = x$enroll_rate,
                      fail_rate = to_sim_pw_surv(fail_rate)$fail_rate,
                      dropout_rate = to_sim_pw_surv(fail_rate)$dropout_rate) |>
            cut_data_by_event(round(x$analysis$event, 0)) |>
[...]

PR #273, opened 2024-08-08
Add info as the output of `sim_gs_n()`

LittleBeannie:
See issue #259 .


keaven (approved):
Looks good

jdblischak:
The title of this PR is confusing. It looks like you added the columns `info` and `info0` *only* to the test function `wlr()`. This has broken the `sim_gs_n()` feature of using different tests per cut. What is the timeline for adding `info` and `info0` columns to `rmst()` and `milestone()`? Can we create an Issue for this?

LittleBeannie:
Adding info and info0 to `rmst` and `milestone` will be a very future plan. I am wondering if `sim_gs_n` could `bind_rows` results from different tests. For example,
image
This `bind_rows` will also help with adding maxcombo (instead of reporting Z score, we report p-values) to `sim_gs_n`.

jdblischak:
Good idea. We can filling missing columns with `NA`s. {dplyr} is not a dependnecy of {simtrial}, but {data.table} is, and it has an equivalent function.


```R
library("dplyr")
library("tibble")
df1 <- tibble(x = 1:2, y = letters[1:2])
df2 <- tibble(x = 4:5, z = 1:2)
r1 <- bind_rows(df1, df2)

rbind(df1, df2)
## Error in match.names(clabs, names(xi)) :
##   names do not match previous names

library("data.table")
r2 <- data.table::rbindlist(list(df1, df2), use.names = TRUE, fill = TRUE)

identical(as.data.frame(r1), as.data.frame(r2))
## [1] TRUE

LittleBeannie:
Shall we create a issue for this?

jdblischak:
> Shall we create a issue for this?

I already have RStudio open, so I'm going to attempt to do it right now

PR #274, opened 2024-08-09
A few minor updates to fix NOTEs from `R CMD check`

jdblischak:
Yesterday I noticed a few NOTEs from `R CMD check`

In R-Devel, there is a new NOTE about missing [package anchors](https://cran.r-project.org/doc/manuals/R-exts.html#Cross_002dreferences-1):

```
* checking Rd cross-references ... NOTE
Found the following Rd file(s) with Rd \link{} targets missing package
anchors:
  rmst.Rd: Surv
Please provide package anchors for all Rd \link{} targets not in the
package itself and the base packages.
```

Since we are using the [roxygen2 markdown syntax](https://roxygen2.r-lib.org/articles/markdown.html), our only option to add the package anchor is `[survival::Surv()]`, which is awkward for the sentence, so I rearranged the link (in direct Rd syntax we could have used `\code{\link[survival]{Surv}}` to add the package anchor while still only displaying `Surv`).

I also updated the global variables that were renamed recently in #267 to fix the following NOTE:

```
  event_total event_trt n_risk_total
```


jdblischak:
Confirmed no more NOTEs from the [R-Devel `R CMD check`](https://github.com/Merck/simtrial/actions/runs/10321932544/job/28575965395?pr=274#step:6:130)

LittleBeannie (approved):
Thanks you, @jdblischak, for cleaning the notes!

PR #276, opened 2024-08-20
Convert wlr() to S3 generic to accept tte_data or counting_process

### LINKED ISSUE #275: Make `wlr` enable a S3 method for both `tte_data` object and `counting_process` object
The `wlr()` function is a S3 method, where we have `wlr.tte_data` and `wlr.counting_process`. 

When the input of `wlr()` is a `tte_data`, we can implement `wlr()` as following. Please add a class `tte_data` to the return of `cut_data_by_event` and `cut_data_by_date`. 
```
x <- sim_pw_surv(n = 300) |> cut_data_by_event(x, 100)
x |> wlr(weight = fh(0, 0.5))
```

When the input of `wlr()` is a `counting_process`, we can implement `wlr()` as following. 
```
x <- sim_pw_surv(n = 20)  |> cut_data_by_event(x, 100) |> counting_process(arm = "experimental")
x |> wlr(weight = fh(0, 0.5))
```

jdblischak:
Closes #275 

Added the class `"tte_data"` to the output of `cut_data_by_event()` and `cut_data_by_date()`, then converted `wlr()` to an S3 generic.  The `"counting_process"` class was already added to the output of `counting_process()` in https://github.com/Merck/simtrial/pull/267


R/wlr.R
```diff
@@ -85,8 +86,34 @@
...
+}
+
+#' @rdname wlr
+#' @export
+wlr.tte_data <- function(data, weight, return_variance = FALSE) {
   x <- data |> counting_process(arm = "experimental")
+  wlr.default(x, weight, return_variance = FALSE)
+}
+
+#' @export
```
LittleBeannie, line 114: Thank you, @jdblischak! I like the idea of `wlr.default` to prevent code duplication. I have a minor comment - should we consider not exporting wlr.default since it is rarely used by users?

jdblischak: `wlr.default` isn't actually exported in `NAMESPACE`. Using `@export` for an S3 method is a roxygen2 convention. It only results in `S3method(wlr,default)` being added to `NAMESPACE`. In other words, the user won't directly see it in the R console, as I demonstrate below:

```R
simtrial::wlr.default
## Error: 'wlr.default' is not an exported object from 'namespace:simtrial'
```

If I remove `@export` from `wlr.default()`, then roxygen2 complains because this would break the S3 dispatch:

```
devtools::document(roclets = c('rd', 'collate', 'namespace'))
## ℹ Updating simtrial documentation
## ℹ Loading simtrial
## ✖ wlr.R:114: S3 method `wlr.default` needs @export or @exportS3method tag.
## Writing 'NAMESPACE'
## Documentation completed
```

jdblischak: However, if `wlr.default()` is distracting to developers, I can remove it. This would require users to always pass `"tte_data"` or `"counting_process"`, which I believe is what we want.

LittleBeannie: Good to know. I am okay with keeping `wlr.default`!

jdblischak:
I rebased onto #273, which has broken my strategy for converting `wlr()` to an S3 generic. The goal was to pre-compute `counting_process()` prior to running multiple iterations of `wlr()`, but now the pre-counting-process data is required to compute the new `ratio`

https://github.com/Merck/simtrial/blob/2ed558b7532373e8846f80923fd7acc98b276854/R/wlr.R#L89-L95

LittleBeannie:
Got you! I guess one possible way is to get `ratio` as an attribute of `data`. 
```
ratio <- sum(data$treatment == "experimental") / sum(data$treatment == "control")
```

jdblischak:
> I guess one possible way is to get `ratio` as an attribute of `data`. 

Ok, I proceeded with that strategy. Though instead of passing the `ratio`, I passed the number of experimental and control treatments.

R/counting_process.R
```diff
@@ -144,6 +147,10 @@ counting_process <- function(x, arm) {
 
   setDF(ans)
   class(ans) <- c("counting_process", class(ans))
+  # Record number of control and experimental treatments, which is required for
+  # downstream test function wlr()
+  attr(ans, "control") <- sum(x$treatment == "control")
```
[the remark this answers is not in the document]
jdblischak: No problem. Done in https://github.com/Merck/simtrial/pull/276/commits/09e9a90facc5190b20033202f0dcbae5cdcb087f

R/wlr.R
jdblischak: I removed `wlr.default()` in https://github.com/Merck/simtrial/pull/276/commits/de83a4515703a7b9cb23e4aad90fd58d3d9088b0. If we decide we want it a default method in the future (eg so that we can pass non-classed objects to `wlr()`), we can always add it then

LittleBeannie (approved):
Thank you, @jdblischak! This PR has been merged.

PR #282, opened 2024-09-13
Add regression tests for summary.simtrial_gs_wlr()

jdblischak:
I am starting to convert `summary.simtrial_gs_wlr()` to data.table (https://github.com/Merck/simtrial/pull/268/files#r1727564684). In order to ensure I don't change the results, I need regression tests.

Issue #271, opened 2024-08-07
Unify the sign of the Z-score for WLR, milestone, and RMST test


LittleBeannie:
For WLR test with MB weights, FH weights, or the early zero weights, the Z-score is a **negative** number:
<img width="553" alt="image" src="https://github.com/user-attachments/assets/b723800c-f949-44a4-b650-c9554a15712b">

While for the RMST test and milestone test, the Z-score is a **positive** number:
<img width="380" alt="image" src="https://github.com/user-attachments/assets/820b6bb0-2356-449a-bc77-4d272010df2d">

The PR https://github.com/Merck/simtrial/pull/272 is to get all the Z-score to positive numbers.

PR #286, opened 2024-09-24
Fix sign of z when passing explicit design to summary.simtrial_gs_wlr()

jdblischak:
Follow-up to #282. In that PR I had only fixed the sign of z for `design = NULL`. However, when an explicit design object is passed to `summary.simtrial_gs_wlr()`, there were multiple additional instances of `-z`.

xref: #271, #272


LittleBeannie (approved):
Thank you, @jdblischak !

Issue #296, opened 2024-11-04
Fix CRAN pre-tests 


LittleBeannie:
```
* using log directory 'd:/RCompile/CRANincoming/R-devel/simtrial.Rcheck'
* using R Under development (unstable) (2024-10-30 r87277 ucrt)
* using platform: x86_64-w64-mingw32
* R was compiled by
    gcc.exe (GCC) 13.2.0
    GNU Fortran (GCC) 13.2.0
* running under: Windows Server 2022 x64 (build 20348)
* using session charset: UTF-8
* checking for file 'simtrial/DESCRIPTION' ... OK
* checking extension type ... Package
* this is package 'simtrial' version '0.4.2'
* package encoding: UTF-8
* checking CRAN incoming feasibility ... [14s] NOTE
Maintainer: 'Yujie Zhao <yujie.zhao@merck.com>'

Unknown, possibly misspelled, fields in DESCRIPTION:
  'Remotes'
* checking package namespace information ... OK
* checking package dependencies ... OK
* checking if this is a source package ... OK
* checking if there is a namespace ... OK
* checking for hidden files and directories ... OK
* checking for portable file names ... OK
* checking whether package 'simtrial' can be installed ... OK
* used C++ compiler: 'g++.exe (GCC) 13.2.0'
* checking installed package size ... OK
* checking package directory ... OK
* checking for future file timestamps ... OK
* checking 'build' directory ... OK
* checking DESCRIPTION meta-information ... OK
* checking top-level files ... OK
* checking for left-over files ... OK
* checking index information ... OK
* checking package subdirectories ... NOTE
Problems with news in 'NEWS.md':
  Cannot extract version info from the following section titles:
    Coding Improvements
* checking code files for non-ASCII characters ... OK
* checking R files for syntax errors ... OK
* checking whether the package can be loaded ... OK
* checking whether the package can be loaded with stated dependencies ... OK
* checking whether the package can be unloaded cleanly ... OK
* checking whether the namespace can be loaded with stated dependencies ... OK
* checking whether the namespace can be unloaded cleanly ... OK
* checking loading without being on the library search path ... OK
* checking use of S3 registration ... OK
* checking dependencies in R code ... OK
* checking S3 generic/method consistency ... OK
* checking replacement functions ... OK
* checking foreign function calls ... OK
* checking R code for possible problems ... [15s] OK
* checking Rd files ... OK
* checking Rd metadata ... OK
* checking Rd line widths ... OK
[...]

PR #262, opened 2024-07-18
Test and document forced evaluation of promises for parallel execution

jdblischak:
Follow-up to PR #261 

I converted the example from Issue #260 into a test. It's a very frustrating test case though. I could only get it to fail pre-#261 when running the code interactively (though it did also fail via `Rscript`). When I ran it via `R CMD check`, it always passed. Unclear why, but I suspect that {testthat} may be responsible.

I also moved the parallel example of `sim_gs_n()` outside of `\dontrun{}`. This was added in https://github.com/Merck/simtrial/commit/749eadcca8d95b2fc3df50653b92274783020c95
(https://github.com/Merck/simtrial/pull/249), and at least when I run `R CMD check` locally, there is no problem with executing this code.

PR #297, opened 2024-11-04
296 fix cran pre tests

LittleBeannie:
To fix issue #296

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -63,6 +63,5 @@ Suggests:
     tidyr
 LinkingTo:
     Rcpp
-Remotes: Merck/gsDesign2
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.3.2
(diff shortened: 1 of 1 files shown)


LittleBeannie:
Hi @jdblischak and @nanxstats, it looks like I only changed the DESCRIPTION file but the CMD check on macos is failed.  Do you happen to have any thoughts?

jdblischak:
We fixed this error in the dev version of {gsDesign2}: https://github.com/Merck/simtrial/pull/262#issuecomment-2237609774

jdblischak (changes requested):
I recommend releasing {gsDesign2} to CRAN (https://github.com/Merck/gsDesign2/pull/480), and then updating the `Imports` of {simtrial} to require this latest version with `>=`

LittleBeannie:
Thanks for the suggestion! I will prepare the release of gsDesign2 first.

PR #299, opened 2024-11-15
Use getFromNamespace instead of ::: to fix S3 dispatch for as_gt()

jdblischak:
Avoid `R CMD check` NOTE about accessing internal functions between packages we maintain (in this case {gsDesign2})

xref: https://github.com/Merck/simtrial/issues/284, https://github.com/Merck/simtrial/pull/287, https://github.com/Merck/gsDesign2/issues/490, https://github.com/Merck/gsDesign2/pull/492

### DIFF
--- a/R/as_gt.R
+++ b/R/as_gt.R
@@ -148,10 +148,12 @@ as_gt.simtrial_gs_wlr <- function(x,
 
 #' @export
 as_gt.fixed_design <- function(x, ...) {
-  gsDesign2:::as_gt.fixed_design(x, ...)
+  f <- utils::getFromNamespace("as_gt.fixed_design", "gsDesign2")
+  f(x, ...)
 }
 
 #' @export
 as_gt.gs_design <- function(x, ...) {
-  gsDesign2:::as_gt.gs_design(x, ...)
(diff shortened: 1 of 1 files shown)


jdblischak:
Demonstration that it still works:

```R
library("gsDesign2")
library("simtrial")
## Attaching package: ‘simtrial’
##
## The following object is masked from ‘package:gsDesign2’:
##
##     as_gt
##

# works without gsDesign2:: prefix
gs_design_ahr() |> summary() |> as_gt()
fixed_design_ahr(
  enroll_rate = define_enroll_rate(duration = 18, rate = 1),
  fail_rate = define_fail_rate(
    duration = c(4, 100),
    fail_rate = log(2) / 12,
    hr = c(1, .6),
    dropout_rate = .001
  )
) |> summary() |> as_gt()
```

jdblischak:
All the builds returned no NOTEs. Though in this case there was no NOTE beforehand, presumably because the dev version of {gsDesign2} is being installed (https://github.com/Merck/simtrial/pull/297). It seems that it is ok with `:::` as long as the internal function exists. Still safer to use `getFromNamespace()` though to avoid any potential problems during CRAN submission

LittleBeannie (approved):
Thank you, @jdblischak !

PR #300, opened 2024-11-15
Requestion gsDesign2 >= v1.1.3

LittleBeannie:
This PR follows #297. 

John might be off today. @nanxstats and @yihui, could you please review and approve it? I will prepare a CRAN submission after your approval.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -63,6 +63,5 @@ Suggests:
     tidyr
 LinkingTo:
     Rcpp
-Remotes: Merck/gsDesign2
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.3.2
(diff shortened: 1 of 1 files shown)


yihui (approved):
I think that's good.

yihui:
Oh actually you *might* not be able to pass the automated check on CRAN because **gsDesign2** 1.1.3 is too new, which may not be available on CRAN's check machines yet, but you can certainly try. If it fails to pass today, you can retry the submission tomorrow and I guess it should be able to pass automatically. Feel free to ping me here if you encounter any problems during the submission.

nanxstats (approved):
Yihui has a valid point but you'll know during CRAN submission. Also, the dependencies in `DESCRIPTION` are sorted alphabetically but that's only a minor style thing.

LittleBeannie:
Thank you, @yihui, for the suggestions!

PR #311, opened 2025-03-04
Document attribute "ratio" returned by counting_process()

jdblischak:
I updated the documentation for `counting_process()` to match its latest behavior. The returned attribute is now `ratio`.

xref: https://github.com/Merck/simtrial/pull/276, https://github.com/Merck/simtrial/pull/281

Idea for future improvement: It would be a good idea to also document the attribute `ratio` in the documentation for `sim_pw_surv()` too. The `counting_process()` code describes the simple ratio of event counts as the "empirical ratio", so I assume that `sim_pw_surv()` must be doing something different to calculate its ratio.

https://github.com/Merck/simtrial/blob/85cdc8fab2cb12d0fc6c36036c7037c1b4995b67/R/counting_process.R#L151-L158


jdblischak:
> I hope this explanation is clear.

@LittleBeannie Yes! Very clear now. Thanks for the detailed explanation. Now I understand this line of code from the end of `sim_pw_surv()`:

https://github.com/Merck/simtrial/blob/289eb5d62912e4392bfc9db28bfd4f41a1047b07/R/sim_pw_surv.R#L195

I'm planning to submit a documentation PR to explain the `ratio` attribute returned by `sim_pw_surv()`

PR #314, opened 2025-03-13
Document sim_pw_surv() calculation of attribute "ratio"

jdblischak:
Follow-up to https://github.com/Merck/simtrial/pull/311#issuecomment-2718898465


LittleBeannie (approved):
I like the value section, thank you for improving the documentation, @jdblischak !

PR #315, opened 2025-03-13
Remove NOTE from unstated dependency in vignette

jdblischak:
Noticed a new NOTE from `R CMD check`:

```
* checking for unstated dependencies in vignettes ... NOTE
'library' or 'require' call not declared from: ‘tibble’
```

This was caused by the line below in the vignette added in https://github.com/Merck/simtrial/pull/308. It can be safely removed.

https://github.com/Merck/simtrial/blob/289eb5d62912e4392bfc9db28bfd4f41a1047b07/vignettes/discrepancy-between-simtrial-and-survival.Rmd#L16


LittleBeannie (approved):
Thank you, @jdblischak !

PR #318, opened 2025-03-25
Replace `foreach(.combine = "rbind")` with manual combine for optimal speed

nanxstats:
This PR replaces `foreach(.combine = "rbind")` with manual combine using `rbindlist()`.

@keaven and I discovered that `foreach(.combine = "rbind")` is an inefficient implementation that can double the compute time compared to combining later manually when the number of simulations gets moderately large (> 10,000) - half of that time is spent on burning one CPU core to combine the data frames.

See my [blog post](https://nanx.me/blog/post/r-foreach-combine/) for generic benchmarking results.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,11 @@
+# simtrial 0.4.2.9000
+
+## Speed improvements
+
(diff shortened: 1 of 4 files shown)


jdblischak (approved):
This is a great improvement. Thanks @nanxstats!

PR #319, opened 2025-04-02
Update parallel vignette to address {data.table} multithreading

jdblischak:
This PR updates the parallel vignette to better compare the results from sequential and parallel processing. We thank @michaelmayer2 for alerting us to the fact that the default multithreading of {data.table} made the sequential processing timings confusing.

I made the following updates:
* Updated the variable names to reflect whether they were created by sequential or parallel processing
* Two important code chunks with the results had set `eval=FALSE`. I replaced the results table with a histogram to demonstrate the difference in durations for the two different enrollment strategies. I used `all.equal()` to demonstrate that `set.seed()` worked as expected
* I set `data.table::setDTthreads(threads = 1)` so that the sequential processing time was completely sequential. I didn't need to do any extra for the parallel processing because {data.table} automatically sets `threads = 1` when it runs inside a forked process. Thus this will behave like we want when run in a non-interactive R session on a Linux machine (namely in GitHub Actions to produce the pkgdown page and on CRAN to produce the bundled vignette). The multithreading behavior gets more complex when run in Windows or from RStudio, but I don't think we need to worry about this

### DIFF
--- a/vignettes/parallel.Rmd
+++ b/vignettes/parallel.Rmd
@@ -31,7 +31,7 @@ The function `sim_fixed_n()` provides a simulation workflow for a two-arm trial 
 We can vary the parameters of the trial using different functions outlined in the documentation.
 This function now provides users the opportunity to implement their simulations using the previously described parallel backends to accelerate the computation.
 
-The function `sim_gs_n()` which simulates group sequential designs under a fixed sample size also supports the use of user-defined backends to parallelize simulations in a similar manner. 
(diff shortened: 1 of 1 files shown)


nanxstats (approved):
I appreciate the suggestions and efforts to improve the correctness and readability of this parallel vignette - it's often a challenge because it's more convoluted than it should be. Now this is the definition of ✨clarity✨

Issue #280, opened 2024-09-05
Enable formula to `wlr()`


LittleBeannie:
The `rmst()` allows formula interface, i.e.,
```
rmst(
  data = ex1_delayed_effect,
  formula = Surv(month, evntd) ~ trt,
  tau = 10,
  reference = "0"
)
```

Shall we enable formula interface to `wlr()`?
```
wlr(
  data = ex1_delayed_effect,
  formula = Surv(month, evntd) ~ trt ,
  weight = fh(0, 0.5)
)
```

If there are strata, what is the best way to save it in the formula?
Keaven's suggestion:
```
wlr(
  data = ex1_delayed_effect,
  formula = Surv(month, evntd) ~ trt + strata(strata variable),
  weight = fh(0, 0.5)
)
```
```

**Comment by LittleBeannie** [comment]
Decision: wait for Larry's weighted cox regression.

**Comment by LittleBeannie** [comment]
A reference R package `coxphw`: https://cran.r-project.org/web/packages/coxphw/index.html

**Comment by jdblischak** [comment]
xref:
  * PRs that added argument `formula` to `rmst()`: #216, #242 
  * PR that converted `wlr()` to an S3 generic: #276
    * If we want to be able to pass arbitrary, unclassed objects to `wlr()`, we need to add back `wlr.default()` (ie revert https://github.com/Merck/simtrial/pull/276/commits/de83a4515703a7b9cb23e4aad90fd58d3d9088b0)

**Comment by jdblischak** [comment]
@LittleBeannie could you please provide some examples that demonstrate the current interface of `wlr()` and then how you would like that to be expressed equivalently with the proposed `formula` argument?

For example, how can I call `wlr()` without using the argument `formula` to generate the same results as the proposed call below?

```R
wlr(
  data = ex1_delayed_effect,
  formula = Surv(month, evntd) ~ trt ,
  weight = fh(0, 0.5)
)
```

`rmst()` has various arguments that begin with `var_`, but `wlr()` doesn't have these, so it isn't immediately clear to me what to do with the variable names after I parse the input formula.

LittleBeannie:
Hi @jdblischak! The following code is equivalent to your example.
```
ex1_delayed_effect |>
  mutate(treatment = trt, tte = month) |>
  wlr(weight = fh(0, 0.5))
```

Please kindly note that, for `wlr.tte_data()`, it requires the `data` to have 4 columns: `tte`, `event`, `stratum` and `treatment`.

[...]

PR #320, opened 2025-04-09
Add wlr.default() to accept custom input data frames

jdblischak:
No reason to require the specific class if they provide the correct column names

xref: making progress towards adding a formula argument (https://github.com/Merck/simtrial/issues/280). This will require `wlr.default()` since the whole point is to accept input data frames whose column names do not match the requirements of the class `tte_data`

background: I had removed `wlr.default()` in https://github.com/Merck/simtrial/commit/de83a4515703a7b9cb23e4aad90fd58d3d9088b0 as part of https://github.com/Merck/simtrial/pull/276 in order to keep it as simple as possible. But at the time I didn't realize we wanted to pass custom input data frames, and requiring the user to manually add the class `"tte_data"` like we did in our example is unnecessarily tedious.


nanxstats (approved):
Looks like a reasonable change

LittleBeannie (approved):
Thanks, @jdblischak!

PR #321, opened 2025-04-10
Minor improvements to `tte_data` class

jdblischak:
Updated a few things I noticed as I am working on #280

PR #322, opened 2025-04-10
Install {future} 1.34.0 for Windows CI

jdblischak:
This is a temporary workaround to fix `R CMD check` on Windows until we find a long-term solution.

**Background:**

* Today I discovered that the `R CMD check` workflow was broken in our Windows CI (https://github.com/Merck/simtrial/pull/321#issuecomment-2794890052)
* This was caused by today's release of {future} 1.40.0
* According to its [`NEWS`](https://cran.r-project.org/web/packages/future/news/news.html): "This update is fully backward compatible with previous versions."
* With {future} 1.40.0, the vignette `parallel.Rmd` fails with `could not find function "sim_pw_surv"`
* I cannot reproduce this error on my local Windows machine. This likely explains why it wasn't caught by the CRAN reverse dependency checks
* My first idea was to add an upper bound of `future (< 1.40)` to `DESCRIPTION`, but for some reason `r-lib/actions/setup-r-dependencies@v2` couldn't process the upper bound on macOS or Windows. It ran fine in the Ubuntu builds. I also confirmed `pak::pak()` could handle the upper bound just fine on my local Windows machine.
* So for now just installing {future} 1.34.0 for the Windows CI until we can figure out what is causing the error

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -44,6 +44,13 @@ jobs:
           extra-packages: any::rcmdcheck
           needs: check
 
+      - name: Install future 1.34 for Windows
+        if: runner.os == 'Windows'
+        shell: Rscript {0}
+        run: |
+          install.packages("remotes")
+          remotes::install_version("future", "1.34.0")
+
       - uses: r-lib/actions/check-r-package@v2
         with:
           upload-snapshots: true


LittleBeannie (approved):

nanxstats:
I also can't reproduce this error on a local Windows PC 😅 

In contrast, I forked the repo and removed your section so it installs future 1.40.0 and [the job](https://github.com/nanxstats/simtrial/actions/runs/14394589490/job/40367898977) reproduces the error.

So, I guess we just need to figure out what's so special about future 1.40.0 + GitHub Actions Windows running environment. It looks like the error actually happens in {doFuture} logic which was updated in Feburary to prepare for this {future} release.

Or maybe we should just ask Henrik on the possible cause with a minimal reproducible example.

jdblischak:
> I also can't reproduce this error on a local Windows PC 😅

Thanks for testing. This further confirms there is something specific to the GitHub Actions environment that is causing the problem.

> Or maybe we should just ask Henrik on the possible cause with a minimal reproducible example.

Exactly. Given that I can't reproduce it locally, I am planning on filing an Issue upstream after I put together a minimal reprex.

jdblischak:
A few updates.

First, I am now able to reproduce the error with {future} 1.40.0 locally. The key is to not have {simtrial} pre-installed (which mimics the Windows CI environment).

```sh
osVersion
## [1] "Windows 11 x64 (build 26100)"
R.version.string
## [1] "R version 4.3.3 (2024-02-29 ucrt)"
remove.packages("simtrial")
## Error in remove.packages : there is no package called ‘simtrial’

# future 1.34.0
remotes::install_version("future", "1.34.0", upgrade = FALSE)
devtools::check()

# Restart R session

# future 1.40.0
remotes::install_version("future", "1.40.0", upgrade = FALSE)
devtools::check()
##
## ─  installing the package to build vignettes
## E  creating vignettes (29.3s)
##    --- re-building 'parallel.Rmd' using rmarkdown
##
##    Quitting from parallel.Rmd:192-214 [confirm-multisession]
##    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
##    <error/rlang_error>
##    Error in `sim_pw_surv()`:
##    ! could not find function "sim_pw_surv"
##    ---
##    Backtrace:
##        ▆
##     1. └─simtrial::sim_fixed_n(...)
##     2.   └─... %dofuture% ...
##     3.     └─doFuture:::doFuture2(foreach, expr, envir = parent.frame(), data = NULL)
##    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
##
##    Error: processing vignette 'parallel.Rmd' failed with diagnostics:
##    could not find function "sim_pw_surv"
##    --- failed re-building 'parallel.Rmd'
##
##    SUMMARY: processing the following file failed:
##      'parallel.Rmd'
##
##    Error: Vignette re-building failed.
##    Execution halted
```

Second, I confirmed the source of the problem is {future} 1.40.0 and not {doFuture} 1.0.2.

| future | doFuture | check |
|--------|--------|--------|
| 1.34.0 | 1.0.2 | pass |
| 1.40.0 | 1.0.2 | fail |
| 1.40.0 | 1.0.1 | fail |

PR #323, opened 2025-04-14
Add argument `formula` to `wlr()`

### LINKED ISSUE #280: Enable formula to `wlr()`

jdblischak:
Closes #280 

xref: #216, #276, #320, #321

This PR adds the argument `formula` to `wlr()`. This allows users to run `wlr()` on their time-to-event data set without having to rename their columns to match the output of one of the cutting functions.

I relaxed `wlr.default()` to not require the column strata. Instead, if it is missing, it adds it automatically for the user.

Also, note that there is the current limitation that the treatment arm must be called `"treatment"`. This is due to this line in `wlr.tte_data()`. In order to make this more flexible, we'd have to add an argument `arm` to `wlr()`.

https://github.com/Merck/simtrial/blob/66de319578154296122cf81136cf4144609b9da0/R/wlr.R#L168


LittleBeannie (approved):
Looks good to me, thanks @jdblischak!

Issue #255, opened 2024-06-03
Provide updated boundary for objects generated by `sim_gs_n`


LittleBeannie:
There are 2 ways to design the API to provide updated bounds:

**API 1:** It added 3 new arguments to the `sim_gs_n` function, which are used to calculate the updated bounds via the `gs_update_ahr` function. 
- The `original_design` is a design object from the gsDesign2 package, which is required when users want to calculate updated bounds. The default is NULL leaving the updated bounds uncalculated.
-The `ia_alpha_spending` is to specify the spend of alpha at interim analysis based on either `"min_planned_actual"` (the minimal of planned and actual alpha spending) or `"actual"` (the actual alpha spending).
- The `fa_alpha_spending` argument is for the scenario when targeted final event count is not achieved (under-running at final analysis): `"info_frac"`  is to spend final alpha according to final information fraction; and `"full_alpha"` is to spend full alpha at final analysis.

Pros: Everything (events, tests, bounds) can be calculated in 1 single parallel set.
Cons: The above API design does not allow users to custom the `ustime` and `lstime`.
```
sim_gs_n(..., 
         original_design= ..., 
         ia_alpha_spending = c("min_planned_actual", "actual"),
         fa_alpha_spending = c("full_alpha", "info_frac"))`
```

**API 2** It keeps the `sim_gs_n` function unchanged and adds a new function `update_bound` to provide updated bounds for both efficacy and futility.

Pros: Users will have the flexibility to customize the `ustime` and `lstime`
Cons: We will first calculate the events, tests in 1 parallel set calling `sim_gs_n`. The updated bounds requires another parallel set calling `update_bound`.
```
# assume the original design is
x <- gs_design_ahr(...)
planned_event <- x$analysis$event

sim_gs_n(...) |> 
   group_by(sim_id) |>
   mutate(ustime = pmin(planned_event, event) / max(planned_event), 
          lstime = ustime) |>
   update_bound(original_design = x)
```

PR #324, opened 2025-04-30
Provide updated boundary for objects generated by `sim_gs_n`

LittleBeannie:
For issue #255.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -53,6 +53,7 @@ Suggests:
     dplyr,
     ggplot2,
     gsDesign,
+    gsDesign2(>=1.1.3.4),
     gt,
     knitr,
     rmarkdown,
@@ -61,8 +62,6 @@ Suggests:
     testthat,
     tibble,
     tidyr
-Remotes:
-    Merck/gsDesign2
 LinkingTo:
     Rcpp
 Roxygen: list(markdown = TRUE)
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -53,6 +53,7 @@ Suggests:
     dplyr,
     ggplot2,
     gsDesign,
+    gsDesign2(>= 1.1.3.4)
     gt,
     knitr,
     rmarkdown,
@@ -61,8 +62,6 @@ Suggests:
     testthat,
     tibble,
     tidyr
-Remotes:
-    Merck/gsDesign2
 LinkingTo:
     Rcpp
 Roxygen: list(markdown = TRUE)
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -53,7 +53,6 @@ Suggests:
     dplyr,
     ggplot2,
     gsDesign,
-    gsDesign2 (>= 1.1.3),
     gt,
     knitr,
     rmarkdown,
@@ -62,6 +61,8 @@ Suggests:
     testthat,
     tibble,
     tidyr
+Remotes:
+    Merck/gsDesign2 (>= 1.1.3.4),
 LinkingTo:
     Rcpp
 Roxygen: list(markdown = TRUE)
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -62,7 +62,7 @@ Suggests:
     tibble,
     tidyr
 Remotes:
-    Merck/gsDesign2 (>= 1.1.3.4),
+    Merck/gsDesign2
 LinkingTo:
     Rcpp
 Roxygen: list(markdown = TRUE)
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -62,6 +62,8 @@ Suggests:
     testthat,
     tibble,
     tidyr
+Remotes:
+    Merck/gsDesign2
 LinkingTo:
     Rcpp
 Roxygen: list(markdown = TRUE)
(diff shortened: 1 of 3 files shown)


DESCRIPTION
```diff
@@ -53,7 +53,6 @@ Suggests:
     dplyr,
     ggplot2,
     gsDesign,
-    gsDesign2 (>= 1.1.3),
```
jdblischak, line 56: This needs to be `gsDesign2 (>= 1.1.3.4),`. The CI jobs are failing because {gsDesign2} is not getting installed (because it isn't in Suggests). You can think of `Remotes` as purely supplemental. It is indicating which of your existing dependencies to obtain from GitHub.

nanxstats: Hadley also has a good [explanation](https://r-pkgs.org/dependencies-in-practice.html#depending-on-the-development-version-of-a-package) of the `Remotes` field in his book.

LittleBeannie: Thank you both! I added `gsDesign2(>= 1.1.3.4)` into the `Suggests` of the DESCRIPTION, and there are no `Remotes`, but the CI jobs failed. Could you please specify if there is a way I could get the CI jobs to run successfully?

jdblischak: > there are no `Remotes`

Why did you remove the section `Remotes`? You need the development version of {gsDesign2} installed from GitHub, right?

To do this, you need _both_ the dependency requirement in `Suggests`:

```
Suggests:
    gsDesign2 (>= 1.1.3.4),
```

_and_ the section `Remotes` to download from the GitHub repository:

```
Remotes:
    Merck/gsDesign2
```

The error you are currently receiving is because you require {gsDesign2} 1.1.3.4, but that is not yet available from CRAN:

```
* deps::.: Can't install dependency gsDesign2 (>= 1.1.3.4)
```

R/sim_gs_n.R
```diff
@@ -38,6 +38,15 @@
...
 #'   examples.
+#' @param original_design A design object from the gsDesign2 package, which is required when users
+#' want to calculate updated bounds. The default is NULL leaving the updated bounds uncalculated.
+#' @param ia_alpha_spending Spend alpha at interim analysis based on
+#' - `"min_planned_actual"`: the minimal of planned and actual alpha spending.
+#' - `"actual"`: the actual alpha spending.
+#' @param fa_alpha_spending If targeted final event count is not achieved (under-running at final analysis),
+#' specify how to do final spending. Generally, this should be specified in analysis plan.
+#' - `"info_frac"` = spend final alpha according to final information fraction
+#' - `"full_alpha"` = spend full alpha at final analysis.
```
jdblischak, line 49: Minor nit: the formatting for the other arguments indents each line with two spaces

R/sim_gs_n.R
```diff
@@ -241,6 +250,21 @@
...
+#'   sample_size = max(x$analysis$n),
+#'   enroll_rate = x$enroll_rate,
+#'   fail_rate = x$fail_rate,
+#'   test = wlr,
+#'   cut = list(ia1 = create_cut(planned_calendar_time = x$analysis$time[1]),
+#'              ia2 = create_cut(planned_calendar_time = x$analysis$time[2]),
+#'              fa = create_cut(planned_calendar_time = x$analysis$time[3])),
+#'   weight = fh(rho = 0, gamma = 0),
+#'   original_design = x
+#' )
```
jdblischak, line 267: We may need to start using `\donttest{}` for some of these examples. The `R CMD check` results identify the `sim_gs_n()`
examples as taking more than 5 seconds:

```R
* checking examples ... [18s/19s] OK
Examples with CPU (user + system) or elapsed time > 5s
          user system elapsed
sim_gs_n 5.016  0.365   6.627
* checking examples with --run-donttest ... [18s/21s] OK
```

This note didn't appear in the [CI runs](https://github.com/Merck/simtrial/actions/runs/14885038375) for the most recent commit to main.

Surprisingly though, when I run this new example on my local machine, it is very fast.

We can wait to address this until immediately prior to the CRAN submission if you like. Especially since we are optimizing the {gsDesign2} code to be faster.

R/sim_gs_n.R
```diff
@@ -330,6 +367,55 @@ sim_gs_n <- function(
...
 
+      # Get event counts per piecewise interval
+      if (!is.null(original_design) && is_logrank){
+        # Get the study duration
+        study_duration <- max(original_design$analysis$time)
+
+        # Get the change point of the piecewise HR, ended by study duration
+        fr_chg_pt <- pmin(cumsum(original_design$fail_rate$duration), study_duration)
+        pw_event <- sapply(fr_chg_pt,
+                           function(threshold) {simu_data_cut |> dplyr::filter(tte <= threshold, event == 1) |> nrow()})
```
jdblischak, line 378: I can convert the {dplyr} calls to either {data.table} or base R after this PR is merged

jdblischak (approved):

R/sim_gs_n.R
LittleBeannie: Thank you, John! I currently reduce the number of simulations `n_sim` to 1. Since the example is just for demo purpose, I guess 1 simulation should be good. If the CRAN asks, we can `dontest{}` this example when do a CRAN release!

R/sim_gs_n.R
LittleBeannie: Thank you! Let's do it later!

PR #325, opened 2025-05-06
Use MathJaX for math rendering in pkgdown >= 2.1.2

nanxstats:
This PR reverts #263 to use the regular MathJax option for math rendering in pkgdown.

Long story short, pkgdown 2.1.2 has fixed a majority of the math rendering problems since pkgdown 2.1.0, where math was not rendered on function reference pages (https://github.com/r-lib/pkgdown/issues/2704), plus MathJax/KaTeX fonts were not properly loaded.

Note that only MathJax and KaTeX options work perfectly in pkgdown 2.1.2. The default MathML option still can't render math properly on function reference pages, and we should avoid using that default option.


keaven:
Thx

jdblischak (approved):

PR #329, opened 2025-05-12
Relax test tolerance for summary.simtrial_gs_wlr()

jdblischak:
Closes #328 

Some minor numerical differences (on the order of `1e-7`) were introduced by https://github.com/Merck/gsDesign2/pull/528. If these can be safely ignored, this PR fixes the tests by relaxing the tolerance to `1e-6`.

PR #330, opened 2025-05-12
Install latest {future} to fix Windows CI

### LINKED ISSUE #327: Does {future} 1.49.0 fix Windows CI issue?
@jdblischak This is to follow up on the workaround added in #322.

Now future 1.49.0 hits CRAN, I wonder if it fixes the check problem introduced by future 1.40.0.

If so, we should revert the GitHub Actions workflow change introduced in #322.

jdblischak:
Closes #327

xref: https://github.com/Merck/simtrial/pull/322, #329 

The bug introduced in {future} 1.40.0 was fixed in 1.49.0

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -44,13 +44,6 @@ jobs:
           extra-packages: any::rcmdcheck
           needs: check
 
-      - name: Install future 1.34 for Windows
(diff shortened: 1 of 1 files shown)


nanxstats (approved):
Great!

LittleBeannie (approved):
Thanks, @jdblischak!

PR #331, opened 2025-05-27
Add test and reduce namespace requirements for `sim_gs_n()` with updated bounds

jdblischak:
This is a follow-up to PR #324 which enabled updating the bounds in a simulation performed by `sim_gs_n()`.

1. First I added a test. I used the deprecated `expect_equivalent()` because we are still using {testthat} edition 2 (the replacement is only available with edition 3)
2. I replaced `dplyr::filter()` with `base::tapply()` and `dplyr::group_by()`+`dplyr::summarize()` with `base::tapply()`. Note that `tapply(summarize = TRUE)` returns a 1-D array, but this appears to be sufficient for our use case

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -54,7 +54,7 @@ Suggests:
     dplyr,
     ggplot2,
     gsDesign,
-    gsDesign2(>= 1.1.3.4),
+    gsDesign2 (>= 1.1.3.4),
     gt,
     knitr,
     rmarkdown,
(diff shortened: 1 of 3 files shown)


nanxstats (commented):
Let's maybe avoid using deprecated testthat APIs when possible? You know we have been using [`local_edition(3)`](https://testthat.r-lib.org/articles/third-edition.html#activating) to switch this at the test level all the time.

jdblischak:
Note that the example I used to create the test does not cover all the code paths. I updated `obs_event`, but only one of its potential downstream manipulations was tested. From the [coverage report of this PR](https://app.codecov.io/gh/Merck/simtrial/pull/331/blob/R/sim_gs_n.R#L394):

image

nanxstats (approved):
The other code paths look familiar... The [gsDesign web interface](https://rinpharma.shinyapps.io/gsdesign/) generates similar code if you switch the "interim and final analysis alpha spending strategy" options in the Update tab.

PR #332, opened 2025-05-29
Reduce runtime of sim_gs_n() examples

jdblischak:
From `R CMD check` on the latest commit https://github.com/Merck/simtrial/commit/99f1aedc3a4a3bacbc2cdb836d42d113833d1946 running on [Ubuntu with R-devel](https://github.com/Merck/simtrial/actions/runs/15328054977/job/43127527403#step:6:116):

```
* checking examples ... [34s/36s] OK
Examples with CPU (user + system) or elapsed time > 5s
           user system elapsed
sim_gs_n 15.684  0.388  17.764
* checking examples with --run-donttest ... [37s/41s] OK
```

To reduce the runtime, I used wrapped most of the tests in `\donttest{}` (which is what we previously did for `sim_fixed_n()` in https://github.com/Merck/simtrial/pull/177).

### DIFF
--- a/R/sim_gs_n.R
+++ b/R/sim_gs_n.R
@@ -137,6 +137,7 @@
 #'   weight = fh(rho = 0, gamma = 0)
 #' )
 #'
+#' \donttest{
 #' # Example 2: weighted logrank test by FH(0, 0.5) at all 3 analyses
 #' sim_gs_n(
 #'   n_sim = 3,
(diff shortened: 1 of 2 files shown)


jdblischak:
Note I had also tried reducing `n_sim` from 3 to 1 so that more code would be run during the package check, but this wasn't sufficient to reduce the runtime of the examples.

LittleBeannie (approved):
Thanks, @jdblischak !

Issue #335, opened 2025-06-09
An example not working for `sim_gs_n`


LittleBeannie:
The following example provides an original design with three analyses. Futility is tested at IA1 only, while efficacy is tested at IA2 and FA only. However, when we simulate this group sequential design using `sim_gs_n`, the current version of simtrial does not work due to a minor data manipulation issue.
```
x <- gs_design_ahr(alpha = 0.025, beta = 0.1,
                   analysis_time = 1:3*12, 
                   upper = gs_spending_bound, upar = list(sf = gsDesign::sfLDOF, total_spend = 0.025),
                   test_upper = c(FALSE, TRUE, TRUE),
                   lower = gs_spending_bound, lpar = list(sf = gsDesign::sfHSD, param = -4, total_spend = 0.01),
                   test_lower = c(TRUE, FALSE, FALSE)
                   ) |> to_integer()

sim_gs_n(
  n_sim = 1,
  sample_size = max(x$analysis$n),
  enroll_rate = x$enroll_rate,
  fail_rate = x$fail_rate,
  test = wlr,
  weight = fh(rho = 0, gamma = 0),
  cut = list(ia1 = create_cut(planned_calendar_time = x$analysis$time[1]),
             ia2 = create_cut(planned_calendar_time = x$analysis$time[2]),
             fa = create_cut(planned_calendar_time = x$analysis$time[3])),
  original_design = x)
```

For a single simulation, the summary data has three rows, with each row corresponding to one analysis. When we attempt to combine the planned/updated bounds with the simulation results, we encounter an issue with the following line:
```
ans_1sim$planned_upper_bound <- x$bound$z[x$bound$bound == "upper"]
```

In this case, `ans_1sim` is a data frame with three rows, while the right-hand side is a vector with only two elements. This issue can be easily resolved using a left join operation:
```
ans_1sim |>
    left_join(original_design$bound |> filter(bound == "upper") |> select(analysis, z) |>  rename(planned_upper_bound = z))
```

PR #336, opened 2025-06-09
Fix the non-working example in `sim_gs_n`

LittleBeannie:
To solve issue #335.


jdblischak (approved):
The macOS error with {future} is unrelated and can need to be fixed in a different PR

Issue #346, opened 2025-09-22
Incorrect cut date per stratum events


LittleBeannie:
The following example discussed the stratified design with 2 strata: biomarker-positive and biomarker-negative.  The cut-off time is decided by the targeted event (50) in the positive population. However, it only gives 21 events. 
```
> sim_data <- sim_pw_surv(n = 360,
                        stratum = data.frame(stratum = c("Positive", "Negative"), p = c(1/3, 2/3)),
                        block = c(rep("control", 2), rep("experimental", 2)),
                        enroll_rate = define_enroll_rate(stratum = c("Positive", "Negative"),
                                                         duration = c(12, 12),
                                                         rate = c(10, 20)), 
                        fail_rate = data.frame(stratum = rep(c("Positive", "Negative"), each = 2),
                                               period = c(1, 1, 1, 1),
                                               treatment = c("control", "experimental", "control", "experimental"),
                                               duration = c(Inf, Inf, Inf, Inf),
                                               rate = log(2) / c(9, 12, 6, 10)),
                        dropout_rate = data.frame(stratum = rep(c("Positive", "Negative"), each = 2),
                                               period = c(1, 1, 1, 1),
                                               treatment = c("control", "experimental", "control", "experimental"),
                                               duration = c(Inf, Inf, Inf, Inf),
                                               rate = c(0.001, 0.001, 0.001, 0.001)
                        ))

> cut_date <- sim_data |> get_analysis_date(target_event_per_stratum = c(50, NA))

> cut_data <- sim_data |> cut_data_by_date(cut_date)

> cut_data |> filter(stratum == "Positive") |> summarize(x = sum(event)) |> pull(x)
[1] 21

> cut_data |> filter(stratum == "Negative") |> summarize(x = sum(event)) |> pull(x)
[1] 50
```

The root cause is here: https://github.com/Merck/simtrial/blob/main/R/get_analysis_date.R#L333. With the `unique()` operation, the stratum value is `c("Negative", "Positive")`, rather than the sequence input by the user, i.e., `c("Positive", "Negative")`.

PR #347, opened 2025-09-22
Correct the cut date when it is driven by the targeted events per stratum 

LittleBeannie:
To solve issue #346.

### DIFF
--- a/R/get_analysis_date.R
+++ b/R/get_analysis_date.R
@@ -318,15 +318,19 @@ get_analysis_date <- function(
 
   # check if target_event_per_stratum is a named vector
   if (!all(is.na(target_event_per_stratum))){
-    if(any(sort(names(target_event_per_stratum)) != sort(unique(data$stratum)))){
-      stop("`target_event_per_stratum` must be a named vector with same values as the data$stratum.")
+    if(is.null(names(target_event_per_stratum))) {
+      stop("`target_event_per_stratum` must be a named vector.")
+    } else if(any(sort(names(target_event_per_stratum)) != sort(unique(data$stratum)))) {
+      stop("`target_event_per_stratum` must be a named vector whose names match the values in data$stratum.")
     }
   }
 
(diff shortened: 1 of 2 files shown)


R/get_analysis_date.R
```diff
@@ -156,7 +156,7 @@
 #' # in each stratum.
 #' get_analysis_date(
 #'   simulated_data,
-#'   target_event_per_stratum = c(100, 200)
+#'   target_event_per_stratum = c("Biomarker-positive" = 100, "Biomarker-negative" = 200)
```
jdblischak, line 159: Should we update the documentation for `target_event_per_stratum` to note that it is required to be a _named_ numeric vector? And should we add a check to `get_analysis_date()` that confirms `names(target_event_per_stratum)` matches the actual strata for the study?

LittleBeannie: Thank you for the suggestion! I updated the documentation and also added the input checking. 
I also updated the `min_n_per_stratum` argument similarly to what I did in `target_event_per_stratum`.

jdblischak: Documentation looks good. Could you please add some tests and then update your code to check for missing names? They don't currently catch the problem because `names()` returns `NULL` when the vector has no names.

```R
example("get_analysis_date")
ls()
get_analysis_date(
  simulated_data,
  target_event_per_stratum = c(100, 200)
)
## Error in .prepareFastSubset(isub = isub, x = x, enclos = parent.frame(),  : 
##   RHS of == is length 0 which is not 1 or nrow (500). For robustness, no recycling is allowed (other than of length 1 RHS). Consider %in% instead.

names(c(100, 200))
## NULL
any(NULL != "b")
## [1] FALSE
```

LittleBeannie: Nice suggestions! I added input checking of named vectors in my function, and some developer tests.

PR #362, opened 2026-05-19
Migrate GitHub Actions to Node 24

jdblischak:
We need to migrate our GitHub Actions to Node 24 by next month

https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
https://github.com/actions/checkout/releases/tag/v6.0.0
https://github.com/codecov/codecov-action/releases/tag/v6.0.0
https://github.com/JamesIves/github-pages-deploy-action/releases/tag/v4.8.0
https://github.com/actions/upload-artifact/releases/tag/v7.0.0

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -29,7 +29,7 @@ jobs:
       R_KEEP_PKG_SOURCE: yes
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
 
       - uses: r-lib/actions/setup-pandoc@v2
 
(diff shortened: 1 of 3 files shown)


yihui (approved):
We can also enable dependabot to automatically bump action versions on a regular basis (e.g. weekly or monthly).

jdblischak:
> We can also enable dependabot to automatically bump action versions on a regular basis (e.g. weekly or monthly).

I agree that is a great suggestion. I made a note to discuss this at the next group meeting since we will need someone with admin access to set this up. My only disagreement is that I think an [interval](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#interval) of "quarterly" is sufficient for our needs.

yihui:
I have no opinion on the frequency.

PR #367, opened 2026-08-24
Migrate to {roxygen2} version 8

jdblischak:


yihui (commented):
The latest roxygen2 version is 8.1.0, and `importFrom()` in `NAMESPACE` will be grouped. Do we want to upgrade to 8.1.0 instead?

jdblischak:
> The latest roxygen2 version is 8.1.0, and `importFrom()` in `NAMESPACE` will be grouped. Do we want to upgrade to 8.1.0 instead?

Good call! I was just trying to get rid of the warning while I was doing some local testing, and I happened to only have 8.0.0 installed. Migrating to the latest 8.1.0 makes more sense. Thanks!

yihui (approved):

PR #368, opened 2026-08-24
Harden summary.simtrial_gs_wlr() tests to upstream {gsDesign2} updates

jdblischak:
My [scheduled integration check](https://github.com/jdblischak/gs-integration-tests/actions/runs/32704703024) identified two {simtrial} tests that fail with the development version of {gsDesign2}. This is because {gsDesign2} was recently updated to change how the futility bound is calculated for the final analysis (https://github.com/Merck/gsDesign2/pull/656).

One of the failing tests is for `sim_gs_n()` and one for `summary.simtrial_gs_wlr()`. This PR only fixes the latter:

```R
── Failure ('test-unvalidated-summary.R:182:3'): summary.simtrial_gs_wlr() returns consistent results for two-sided design ──
Expected `observed` to equal `expected`.
Differences:
actual vs expected
                asy_lower_prob
  actual[1, ]     0.0007883884
  actual[2, ]     0.0570706442
- actual[3, ]     0.1000001794
+ expected[3, ]   0.1000401801

  `actual$asy_lower_prob`: 0.0007884 0.0570706 0.1000002
`expected$asy_lower_prob`: 0.0007884 0.0570706 0.1000402
```

In general, I tried to make these tests more robust to upstream changes in {gsDesign2} by always pulling the expected values for the asymptotic results directly from the design object.

PR #369, opened 2026-08-28
Harden sim_gs_n() tests to upstream {gsDesign2} updates

jdblischak:
Follow-up to #368. This PR updates the test of the `sim_gs_n()` argument `original_design` to be robust to upstream changes in the behavior of {gsDesign2} (in this case, the change in the calculation of the futility bound for the final analysis in https://github.com/Merck/gsDesign2/pull/656). Instead of requiring exact values, it only confirms that 1) the planned bounds match the original design object, and 2) the updated bounds are different from the planned bounds.


jdblischak:
I also confirmed locally that the test passes with either the CRAN or GitHub version of {gsDesign2}

LittleBeannie (approved):

PR #370, opened 2026-08-31
Remove archived {survMisc} from Suggests

jdblischak:
Close #359 

* The {survMisc} package continues to be archived on [CRAN](https://cran.r-project.org/package=survMisc), and no development has occurred in the upstream [repository](https://github.com/dardisco/survMisc) in years
* Our previous fix focused on the immediate problem of restoring our GitHub Actions pipelines (#361)
* But we will be unable to submit to CRAN as long as we have an archived package listed in Suggests, which I confirmed by submitting to winbuilder:
    ```
    Suggests or Enhances not in mainstream repositories:
      survMisc
    ```
* This PR attempts to enable a future CRAN submission while continuing to run the existing tests that require {survMisc} by using the `Config/Needs` approach of [{remotes}](https://remotes.r-lib.org/news/index.html?q=config/needs#new-functions-and-features-2-2-0), which is supported by [r-lib/actions/setup-r-dependencies](https://github.com/r-lib/actions/tree/v2-branch/setup-r-dependencies)
* I confirmed in GitHub Actions run on my fork that {survMisc} is installed, eg [here](https://github.com/jdblischak/simtrial/actions/runs/33411288338/job/99551202260#step:5:13993)

Thoughts on this approach? Does `Config/Needs/tests` work or should we choose a different designation?

Issue #359, opened 2026-05-18
CRAN package {survMisc} was archived


jdblischak:
The CRAN package [{survMisc}](https://cran.r-project.org/package=survMisc) was archived:

<img width="800" height="213" alt="Image" src="https://github.com/user-attachments/assets/6b14d35a-e5a5-4835-92ad-6d9b63cb7087" />

This causes the {simtrial} CI jobs to all fail:

```
Error: 
  ! error in pak subprocess
  Caused by error: 
  ! Could not solve package dependencies:
  * deps::.: Can't install dependency survMisc
  * survMisc: Can't find package called survMisc.
```

{survMisc} is in Suggests because it is used in a few independent tests in [tests/testthat/test-independent_test_fh_weight.R](https://github.com/Merck/simtrial/blob/0e1efcae09a567c2480d7fa6d0c5108277b3c949/tests/testthat/test-independent_test_fh_weight.R#L2) and [ests/testthat/test-independent_test_pvalue_maxcombo.R](https://github.com/Merck/simtrial/blob/0e1efcae09a567c2480d7fa6d0c5108277b3c949/tests/testthat/test-independent_test_pvalue_maxcombo.R#L2)

https://github.com/Merck/simtrial/blob/87748f296b17efb3405705e448c81ae894829fa1/DESCRIPTION#L62

jdblischak:
The NOTEs from `R CMD check` for {survMisc} are easy fixes, but a PR was opened last year to fix one of them, and it is still open (https://github.com/dardisco/survMisc/pull/26)

jdblischak:
We implemented a workaround to fix CI in #361. Longer-term we need to decide how we want to maintain these tests. I doubt that the CRAN maintainers will allow us to submit with an archived package listed in Suggests.

PR #377, opened 2026-09-14
Only install {survMisc} when calculating test coverage

jdblischak:
This is a minor follow-up to my previous PR #370 that dealt with the archived package {survMisc} (#359). In that PR I used the `Config/Needs` strategy to install {survMisc} only in our CI runs. The goal was to continue to run these tests during development while avoiding any problems with CRAN policies.

However, while I had confirmed that {survMisc} was installed, I had missed that the tests were still skipped in the `R CMD check` workflow. After some troubleshooting both locally and in GitHub Actions, it is my understanding that this is a fundamental limitation of `R CMD check`, and not a shortcoming of eg `r-lib/actions/check-r-package`. From what I can tell, `R CMD check` creates a temporary directory that only contains packages that are explicitly listed in `Imports`/`Suggests`/etc. Thus even though {survMisc} was installed, it was not available to {testthat} running inside of `R CMD check`.

However, since our coverage pipeline directly runs `covr::package_coverage()`, it does run those tests, which I feel is sufficient. To make it clear that the {survMisc}-related tests are only being run in the coverage workflow, I have updated `DESCRIPTION` and the workflow files accordingly.


jdblischak:
{testthat} output in test coverage [log](https://github.com/Merck/simtrial/actions/runs/34881505458/job/104101653210?pr=377#step:7:63) showing zero skipped tests:

```
[ FAIL 0 | WARN 1 | SKIP 0 | PASS 275 ]
```

yihui (approved):
> From what I can tell, `R CMD check` creates a temporary directory that only contains packages that are explicitly listed in `Imports`/`Suggests`/etc.

Yes, that's true.

